PR #19, opened 2018-04-11
BCN - Pere

Interna1ta:
Done with the team :)

We managed to connect to the API but not to manage it.... so we created fake data.


.eslintrc.json
```diff
@@ -0,0 +1,22 @@
...
+        "indent": [
+            2,
+            2
+        ],
+        "semi": [
+            "error",
+            "always"
+        ]
+    }
+}
```
tawebbcn: Eslint :+1: :+1:

PR #21, opened 2018-04-11
PAR Clément

Kleemdub:


ta-web-paris:
Good job ! 💯 

Only thing I can add, is that you can use another way to pass the data you receive from the API call, you can use it like this : 

```js

 let beersList = beers;
      res.render("beers", { beersList });

```

Anyway you will see it later, keep up the good work Clément ! 

Arthur/

PR #22, opened 2018-04-11
BCN - Jorge

jorgeterol:


starter_code/app.js
```diff
@@ -1,18 +1,31 @@
...
 
 app.get('/', (req, res, next) => {
   res.render('index');
 });
 
+app.get('/beers', (req, res, next) => {
+  punkAPI.getBeers()
+    .then(beers => {
+      const beerArray = [];
+      beers.forEach(beer => beerArray.push(beer.name));
```
tawebbcn: You didn't had to do this here. In the partial you could also choose which data to use.

PR #41, opened 2018-04-12
BCN - Jordi

jordicalzada:
10% done


starter_code/app.js
```diff
@@ -6,13 +6,37 @@ const path    = require('path');
...
+//config app
+
 app.set('view engine', 'hbs');
 app.set('views', __dirname + '/views');
 app.use(express.static(path.join(__dirname, 'public')));
 
+// route
+
 app.get('/', (req, res, next) => {
   res.render('index');
```
tawebbcn: nice, now you have to improve that

tawebbcn: It finally should look like:
```
  punkAPI.getBeers()
    .then(beersResults => {
      const data = {beersResults};
      res.render('beers', data);
    })
    .catch(error => {
      console.log(error);
  });
});
```

PR #42, opened 2018-04-12
BCN Marketa - Iteration 1,2

Hrabenka7:


starter_code/app.js
```diff
@@ -14,5 +14,14 @@ app.get('/', (req, res, next) => {
   res.render('index');
 });
 
+app.get("/beers", (req, res) => {
+  res.render("beers");
```
tawebbcn: nice, now you have to improve that

tawebbcn: It finally should look like:
```
  punkAPI.getBeers()
    .then(beersResults => {
      const data = {beersResults};
      res.render('beers', data);
    })
    .catch(error => {
      console.log(error);
  });
});
```

starter_code/views/layout.hbs
```diff
@@ -0,0 +1,28 @@
...
+<body>
+    <header id="site-header">
+        <nav class= "nav-bar">
+                <a href="/">Home</a>
+                <a href="beers">Beers</a>
+                <a href="random-beer">Random Beer</a>
+        </nav>
+    </header>
+
+    {{{body}}}
```
tawebbcn: :+1:

PR #45, opened 2018-04-12
PAR Renaud

Ayss83:


starter_code/app.js
```diff
@@ -7,11 +7,67 @@ const PunkAPIWrapper = require('punkapi-javascript-wrapper');
...
 app.set('views', __dirname + '/views');
-app.use(express.static(path.join(__dirname, 'public')));
+
+app.use(express.static(path.join(__dirname, '/public')));
 
 app.get('/', (req, res, next) => {
   res.render('index');
+}); 
+
+app.get("/beers", (req, res, next) => {
```
ta-web-paris: It works but there is an easier way to write (I mean really ;) ), as the data you are receiving from the API call are already inside an array (try to console log it :) ).

You can use this piece of code : 

```js 

app.get("/beers", (req, res, next) => {
  punkAPI
    .getBeers()
    .then(beers => {
      let beer = beers;
      res.render("beers", { beer });
    })
    .catch(error => {
      console.log(error);
    });
});

```

ta-web-paris:
Good job, you can have the same result with less code (which is cool, trust me :p )

Keep up the good work Renaud 🥇

PR #47, opened 2018-04-12
PAR Billy

Yaats:


ta-web-paris:
It looks like that you had some trouble with your partial, I couldn't figure where it came from so I rewrite a piece of code and it was working, any way it looks like you understand how to use the dynamic rendering, but you need to pay more attention to details ;)

If you have trouble or if you need any advice let me know 🥇 

Arthur/

PR #48, opened 2018-04-12
PAR Bertrand

wavilo:
not a lot of time for the CSS
random beer is not done


ta-web-paris:
It's okay, it looks like you understand how to use and play with Express dynamic rendering!

If you need any advice just let me know :)

Arthur/

PR #51, opened 2018-04-12
PAR Patrycja

kucharskap:
Skipped come css in order to try to make the random work, but it still doesnt

### DIFF
--- a/starter_code/app.js
+++ b/starter_code/app.js
@@ -10,9 +10,30 @@ app.set('view engine', 'hbs');
 app.set('views', __dirname + '/views');
 app.use(express.static(path.join(__dirname, 'public')));
 
+hbs.registerPartials(__dirname + '/views/partials');
+
 app.get('/', (req, res, next) => {
   res.render('index');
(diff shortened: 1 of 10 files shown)


starter_code/app.js
```diff
@@ -10,9 +10,30 @@ app.set('view engine', 'hbs');
...
+    punkAPI.getBeers()
+    .then(beers => {
+      res.render('beers', {beers});
+    })
+    .catch(error => {
+      console.log(error);
+    })   
+});
+
+app.get('\random-beers', (req, res, next) =>{
```
ta-web-paris: Whoops! You used a backslash here, this should be `/random-beers`. This is why the route handler is not matching and you are getting a 404

starter_code/views/random-beers.hbs
```diff
@@ -0,0 +1,3 @@
+<div class="container">
+{{>randomBeer this}}
```
ta-web-paris: Great use of partials

PR #52, opened 2018-04-12
PAR Elise

elisegriset92:


starter_code/app.js
```diff
@@ -1,18 +1,46 @@
...
+    })
+    .catch(error => {
+      console.log(error);
+    });
+});
+
+// ---------------
+
+app.get('/random-beers', (request, response, next) => {
+  response.render('beers.hbs');
```
ta-web-paris: This is missing the logic of fetching the random beer and passing it as data to the view.

elisegriset92:
Hey Mickael, thanks for the review.

I didn't have time for the rest of the exercise!
About the node modules, it is not suppose to be send when i use .gitignore?

PR #53, opened 2018-04-12
PAR Maggie

mbbaber:
had some trouble with CSS, so sit looks ugly, but otherwise, I understood the concepts of this exercise.

### DIFF
--- a/starter_code/views/random-beers.hbs
+++ b/starter_code/views/random-beers.hbs
@@ -1,0 +1,17 @@
+<!-- On the randomBeer.hbs you should print the random beer you get. 
+You should display an image, name, description, tagline, food pairing and brewer tips -->
+
+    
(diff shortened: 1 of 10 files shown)


starter_code/views/random-beers.hbs
```diff
@@ -0,0 +1,17 @@
+<!-- On the randomBeer.hbs you should print the random beer you get. 
+You should display an image, name, description, tagline, food pairing and brewer tips -->
+
+    
+<div class="cards-container">
+    <div class="card" style="width: 18rem;">
+      <img class="card-img-top" src= ""{{beerRandom.image_url}}"" alt="Card image cap">
```
ta-web-paris: I think you have a syntax error with the ` "" ` in your `src` attribute, which is causing the beer image to not load properly

PR #54, opened 2018-04-12
PAR Brianna

BrianNacho:
few things:
1. Seems like my commit was larger than expected - not sure what I did wrong
2. Can't make css work for beers page


starter_code/views/randomBeer.hbs
```diff
@@ -0,0 +1,22 @@
+<div class="random-beer">
+  {{#with randombeer}}
+    <div class="row">
+      <div class="col-md-3">
+            
+        <img src="{{image_url}}" width=80 />
+
+      </div>
+
+      <div class="col-md-7">
```
ta-web-paris: You could use your beer partial here as well to make it more DRY :)

PR #56, opened 2018-04-12
PAR JB

Jbmouterde:
Haven't use the partial file but it works for booth beers & random beer


starter_code/app.js
```diff
@@ -13,6 +13,34 @@ app.use(express.static(path.join(__dirname, 'public')));
 app.get('/', (req, res, next) => {
   res.render('index');
 });
+//partial 
+hbs.registerPartials(__dirname + '/views/partials');
+
+app.get('/beers', (req, res, next) => {
+  punkAPI.getBeers()
+  .then(beers => {
+    res.locals.beer= beers
```
ta-web-paris: a better name for the local var would be `beers` since it's an array :)

PR #60, opened 2018-04-12
PAR Mostafa Gad

M-Gadd:


app.js
```diff
@@ -0,0 +1,49 @@
...
+});
+
+app.get("/beers", (req, res, next) => {
+  // const index = Math.floor(Math.random()*25);
+  // NOT SURE---------------
+  punkAPI.getBeers()
+  .then(beers => {
+    console.log(beers)
+    const chosBeer = beers;
+    res.render("beers.hbs", {chosBeer});
```
ta-web-paris: nice use of shorthand object syntax

app.js
```diff
@@ -0,0 +1,49 @@
...
+  })
+  .catch(error => {
+    console.log(error)
+  })
+});
+
+app.get("/random-beers", (req, res, next) => {
+  punkAPI.getRandom()
+  .then(beers => {
+    const chosBeer2 = beers[0];
```
ta-web-paris: you don't have to name it `chosBeer2` since it's in the local scope of this callback, so you can safely name it like the former one

views/random-beers.hbs
```diff
@@ -0,0 +1,10 @@
+<div>
```
ta-web-paris: You could use the partial here as well and make it even more DRY

PR #61, opened 2018-04-12
Not there yet! but will get there for sure!

LeoBethel:


starter_code/app.js
```diff
@@ -14,5 +14,16 @@ app.get('/', (req, res, next) => {
   res.render('index');
 });
 
+hbs.registerPartials(__dirname + "/views/partials");
+
+punkAPI.getBeers().then(beers => {
```
ta-web-paris: This should be inside a route handler (`app.get.... `)

PR #67, opened 2018-04-25
MIA-FTWD-DiegoLeal-Ironbeers

DiegoLealB:


MarcosPalaciosIII:
excellent job on the beers page. Really nice css especially utilizing auto width to set the height of the images in order to get them looking the same.   

Keep up the excellent work.

PR #68, opened 2018-04-25
MIA-FTWD-DALIROJAS-IRONBEERS

rojasdali:


starter_code/app.js
```diff
@@ -15,4 +16,38 @@ app.get('/', (req, res, next) => {
 });
 
 
+app.get('/beers', (req, res, next) => {
+   let data = {
```
sandrabosk: I don't see the reason to use data thing here. You could've skipped this step completely.

starter_code/app.js
```diff
@@ -15,4 +16,38 @@ app.get('/', (req, res, next) => {
...
+  
+}
+  
+});
+
+
+
+
+app.get('/random-beers', (req, res, next) => {
+  let data ={
```
sandrabosk: The same as above ^

sandrabosk:
Goo job. Thank you for submitting.

PR #71, opened 2018-04-25
MIA - FTWD - jose arjona - IRONBEERS LAB

josedarjona:


starter_code/views/partials/randomBeerPartials.hbs
```diff
@@ -0,0 +1,31 @@
...
+    </div>
+    <div class="col-md-9">
+      <div class="card-body">
+        <h2 class="card-title">{{this.name}}</h2>
+        <h5 class="card-title text-muted">{{this.tagline}}</h5>
+        <p class="card-text">{{this.description}}</p>
+        <h5>Food Pairings</h5>
+        <ul>
+          {{#each this.food_pairing}}
+          <li>{{this}}</li>
```
sandrabosk: Great! Although we are working with only one object so each is not needed as well as the partial, it's good for practicing.

sandrabosk:
Good work. Thank you for submitting.

PR #86, opened 2018-04-27
MIA-FTWD-Ruben-Vargas-IronBeers

Ambition101:


sandrabosk:
All good. The only thing I would change is the styling of the all beers page, it looks super confusing having the beer image as background image. Thank you for submitting.

PR #98, opened 2018-06-13
BER -spela

spelavrtovec:


ta-web-ber:
HI Spela,

I think you lacked of times.
We gonna set a catch up session on this exercice.
In my opinion it could be nice if you can join us.
This exercice is very useful, and if you masterize IronBeers, all the future exercices would be easy.

Let me know

#Vivian

PR #102, opened 2018-06-14
BER - Elliot _ Drug Addiction Problems in Ironhack *notfunny*

LJSkadi:
…an addiction rehab


starter_code/app.js
```diff
@@ -9,10 +9,40 @@ const punkAPI = new PunkAPIWrapper();
...
+  .catch(error => {
+    console.log(error)
+  })
+
+  
+});
+
+//Route randombeer
+app.get('/random-beer', (req, res, next) => {
+  const randombeer= punkAPI.randomBeer()
```
ta-web-ber: The method is:
getRandom();

In other hand, you can clean your code like this:
```js
app.get("/random-beer", (req, res, next) => {
	punkAPI
		.getRandom()
		.then(randombeer => {
			res.render("random-beer", randombeer);
		})
		.catch(error => {
			console.log(error);
		});
});
```

PR #109, opened 2018-06-18
MIA PT WD - Maria Quezada

mafeq03:
Basics done - Working on styling of beers and random beer


ta-web-mia:
Great It works as it should which is the most important. Styling can take a while so no worries  n_n

-Adrian

PR #111, opened 2018-06-21
PAR Ghaith, Jana &Shanshan

moonchie:
@janamr 
@GhaithHan
We finished the Express. We didn't do lots of styling. Our ideal was to use Bootstrap and we got the div and CDN.

### DIFF
--- a/starter_code/.gitignore
+++ b/starter_code/.gitignore
@@ -1,0 +1,1 @@
+node_modules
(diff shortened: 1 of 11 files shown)


ta-web-paris:
Exercise well understood!
Just try to watch your code indentation, it really improves readability and helps debugging when something doesn't work!
Keep up the good work :)

PR #174, opened 2018-07-05
BCN-lab-ironbeers

Lilhak4:
Had trouble with the server. Frustrated. Will practice over the weekend.


starter_code/app.js
```diff
@@ -10,9 +10,28 @@ app.set('view engine', 'hbs');
 app.set('views', __dirname + '/views');
 app.use(express.static(path.join(__dirname, 'public')));
 
+app.set('beers', __dirname + '/beers');
```
tawebbcn: you just need to set the views folder, not each view

PR #175, opened 2018-07-05
BCN - ironbeers - just the first 3 iterations done - Julien 

Julienchilhagopian:
Didn't manage to use the API :(

### DIFF
--- a/starter_code/app.js
+++ b/starter_code/app.js
@@ -1,18 +1,31 @@
 
-const express = require('express');
-const hbs     = require('hbs');
-const app     = express();
-const path    = require('path');
-const PunkAPIWrapper = require('punkapi-javascript-wrapper');
-const punkAPI = new PunkAPIWrapper();
+const express = require('express')
+const hbs = require('hbs')
+const app = express()
+const path = require('path')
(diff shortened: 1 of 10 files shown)


starter_code/views/beers.hbs
```diff
@@ -0,0 +1,5 @@
+<div class="beer-container">
+{{#each beers}} 
+{{ }}
```
tawebbcn: and here you can add a partial
```{{> beerPartial this}}```

remember that hen you have to register the partial

File: starter_code/app.js [hunk 2]
```diff
@@ -1,18 +1,31 @@
...
 app.get('/', (req, res, next) => {
-  res.render('index');
-});
+  res.render('index')
+})
 
+app.get('/beers', (req, res, next) => {
+  const beers = punkAPI.getBeers()
+  beers.then(beer => {
+    res.render('beers')
```
tawebbcn: you have to pass data here, and also need a bit of refactoring (like the ```beers``` in line 19):
```
  punkAPI.getBeers()
  .then(beers => {
    res.render('beers', { beers });
  })
});
```

PR #181, opened 2018-07-05
MIA Ironhack Labs - Ironbeers - Ruben

J4CODE:


starter_code/views/randombeer.hbs
```diff
@@ -0,0 +1,5 @@
+<div class="index-container">
```
MarcosPalaciosIII: this page is same as the index page. Should be displaying information for 1 Random Beer. You could have done something like this 
`
<h1> Random Beer </h1>

{{> beerPartials beer}}

`

PR #203, opened 2018-07-09
MIA Hannah Loop

hloop:
Needs to be styled.


starter_code/app.js
```diff
@@ -9,10 +8,30 @@ const punkAPI = new PunkAPIWrapper();
...
+    res.render('beers', {beersArray: blah});
+    })
+  });
+
+
+  // Why is it random-beer and not randomBeer?
+  app.get('/random-beer', (req, res, next) => {
+    punkAPI.getRandom()
+    .then(beers => {
+      res.render('beers', {theBeer:beers[0]});
```
MarcosPalaciosIII: you should be rendering randomBeer instead of beers

starter_code/views/randomBeer.hbs
```diff
@@ -0,0 +1,3 @@
+<h1>Random Beer</h1>
+
+{{> _singleBeer theBeer}}
```
MarcosPalaciosIII: the partial name should be the same as the name of the partial folder: _singleBeer should be singleBeer

MarcosPalaciosIII:
great job on the ironbeers lab. Just remember the small details like the page your going to render or naming of what your using since this can lead to features not working as intended.

keep up the great work

PR #250, opened 2018-08-30
john 

johndimitroff:


ta-web-paris:
This is very good! :)
the data is properly passed to the beers.hbs, the loop is clean and you have all the beer info you needed.
The randomBeer iteration works on the same mechanics so I have no doubt you would have nailed it too

Good job!
Lucas

PR #251, opened 2018-08-30
PAR Abi & Pauline

paulettepaillette:
@abisummers


app.js
```diff
@@ -0,0 +1,50 @@
...
+    .catch(error => {
+      console.log(error)
+    })
+});
+
+app.get('/random-beers', (req, res, next) => {
+  punkAPI.getRandom()
+    .then(beers => {
+      const randomIndex = Math.floor(Math.random() * beers.length);
+      res.locals.randomBeer = beers[randomIndex];
```
ta-web-paris: I see what you tried to do here. It actually works, but it's unnecessary complexity: let me explain.

The .getBeers() method returns an object containing an array of beers. If you wanted to find a random beer inside this array, you would indeed have had to define a random index in this array.

The .getRandom() method, though, doesn't return an array of many objects, but an array of only one, random item. Thus, what you're doing is each time it gives you one item, you randomly find a number between 0 and 0.99999, Math.floor it to 0 and target the only item in the "beers" array. But you could also simply pass `beers[0]` and have your random beer every time anyway :)

This is why I always stress out that checking what you get from your API / backend calls is very important: it actually lets you know what data you're dealing dealing with, and can save you some precious time and effort :)

ta-web-paris:
This is very good! :)
The data is properly passed to the hbs files, the loops are clean, you have all the beer info you needed and the partial looks good.


Good job!
Lucas

PR #343, opened 2018-11-09
[WEB-PT-MAD02] Ferran Lao

FerranLao:


starter_code/views/index.hbs
```diff
@@ -1,3 +1,5 @@
 <div class="index-container">
-  
+   <img  src="/images/beer.png">
+   <button href="/beers" >Check the beers</button>
```
ta-web-mad: This buttons are not working :( 
If you want that a button became a link, has to be inside an <a> or contains an <a>, button itself doesn't have an href attribute.

ta-web-mad:
Good work Ferran! :)

#closed #checked by Cami

PR #381, opened 2018-11-18
MIA FTWD Javier Arrinda

javierarrinda:


starter_code/views/beers.hbs
```diff
@@ -0,0 +1,8 @@
+<div class="theBeerList">
+{{#each beers}}
+    {{> beers this}}
```
ta-web-mia: Instead of beers, it should be the name of the partial file. In your case is beerPartial

starter_code/views/random-beers.hbs
```diff
@@ -0,0 +1,13 @@
+<div class="theRandomBeerBody">
+    {{#each oneBeer}}
+    {{> beerPartial oneBeer}}
```
ta-web-mia: instead of 'oneBeer' write 'this' which would reference to the 'oneBeer' in the each

ta-web-mia:
Good Job!!!

PR #401, opened 2018-11-22
[WEB PT Paris] Sandra & Rita

mernri:
@ta-web-paris 
@sanddpt


starter_code/views/beer.hbs
```diff
@@ -0,0 +1,15 @@
+<div class="cards-container">
+    {{!-- on appelle le contenu du partial beers --}}
+    {{#each beers}}
```
ta-webpt-par: This file is called `beer` but it iterates over a beer array. The server should have passed the only beer object from the array to this view.

This inconsistency comes from the API, when you call he function `getRandom()` it returns an array containing the random beer. Since it is the server that calls this API, it is it's job to format the returned data.

ta-webpt-par:
Great job !

PR #402, opened 2018-11-22
[paris - nico_prieur] lab-ironbeers

nicolasprieur:


starter_code/views/beers.hbs
```diff
@@ -0,0 +1,7 @@
+<ul>
+  {{#each beers}}
+      {{>beers_partial }} //calling partial
+  {{else}}
+    <p>No beers yets!</p>
```
ta-webpt-par: :+1:

starter_code/app.js
```diff
@@ -9,13 +9,36 @@ const punkAPI = new PunkAPIWrapper();
...
+   .catch(error => {
+     console.log(error)
+   })
+});
 
+app.get('/random-beer', (req, res, next) => {
+  punkAPI.getRandom()
+    .then(beers => {
+      // res.send(beers)
+      res.render('randomBeer', {beers});
```
ta-webpt-par: The API is wrongly coded, it should not return an array of beers containing only one random beer, it should return only the random beer. This problem should be resolved by the server, not by the views. You should render the view like this: `res.render('randomBeer', beers[0]);` and you should not iterate over a beer array in the randomBeer view.

ta-webpt-par:
Great work !

PR #441, opened 2019-01-26
[PTMIA1018] Gustavo Galvez

g-galvez2018:
Beers view may need styles to be updated, but everything else seems to work fine.


ta-web-mia:
Well done! Just a tip for universality - <button></button> doesn't work on firefox. Using <input type="button"> is a good alternative that'll work on most if not all browsers
- Francisco

PR #448, opened 2019-01-30
BCN-MiguelNeto-WDFT0119

MR-Neto:


views/layout.ejs
```diff
@@ -0,0 +1,28 @@
...
+    <header>
+        <a href="/">Home</a>
+        <a href="/beers">Beers</a>
+        <a href="/randomBeers">Random Beer</a>
+    </header>
+    <div class="container">
+        <%- body %>
+    </div>
+    <footer>
+        Developed by Ironhack
```
axelgar: I'm guessing this should be a p tag

PR #459, opened 2019-01-31
BCN - Nicolas Lacaze - WDFT0119

nicolacaze:


starter_code/app.js
```diff
@@ -1,21 +1,41 @@
 
 const express = require('express');
-const hbs     = require('hbs');
-const app     = express();
 const path    = require('path');
+const expressLayouts = require('express-ejs-layouts');
 const PunkAPIWrapper = require('punkapi-javascript-wrapper');
 const punkAPI = new PunkAPIWrapper();
```
axelgar: Try to separate the requires from other variables that you add. It's a way to keep your code more structure.

starter_code/public/stylesheets/styles.css
```diff
@@ -5,4 +5,159 @@ dark-blue: #0A2C42;
 grey: rgba(145, 145, 145, 0.481);
 font-family: Arial, Helvetica, sans-serif
 */
+
```
axelgar: Good that you used the boilerplate, but if there are things you don't use, it's better if you delete them.

starter_code/app.js
nicolacaze: Noted!

starter_code/public/stylesheets/styles.css
nicolacaze: Gotcha 🧐

PR #460, opened 2019-01-31
BCN Ranieri Galasso

ranierigalasso:


starter_code/app.js
```diff
@@ -1,21 +1,48 @@
 
 const express = require('express');
-const hbs     = require('hbs');
+const ejs     = require('ejs');
 const app     = express();
```
axelgar: Try to put all the requires at the top separated from other variables that you add. You will get a cleaner code.

starter_code/app.js
```diff
@@ -1,21 +1,48 @@
...
 const punkAPI = new PunkAPIWrapper();
+const expressLayouts = require('express-ejs-layouts');
 
-app.set('view engine', 'hbs');
+app.use(expressLayouts);
+app.set('layout', 'layout');
+app.set('view engine', 'ejs');
 app.set('views', __dirname + '/views');
 app.use(express.static(path.join(__dirname, 'public')));
+app.use(express.static('public/stylesheets'));
```
axelgar: I'm not sure you need these two lines of code. With the line 14 should be enough.

starter_code/public/stylesheets/styles.css
```diff
@@ -5,4 +5,114 @@ dark-blue: #0A2C42;
...
+  width: auto;
+}
+.random-image {
+  margin: 0 10%;
+}
+.random-beer p {
+  margin: 1%;
+}
+.random-beer h1 {
+  margin: 1%;
```
axelgar: Maybe there is a way to fix this margin thing without having to create so many rules?

PR #473, opened 2019-01-31
BCN-YOSHITSUGU NAGAO-WDFT19

YoshitsuguNagao:


starter_code/app.js
```diff
@@ -1,21 +1,53 @@
...
+app.set('views', path.join(__dirname, '/views'));
+app.use(express.static(path.join(__dirname, 'public')));
 
+app.use(expressLayouts);
+app.set('layout', 'layout');
 
-app.get('/', (req, res, next) => {
+// app.get('/', (req, res, next) => {
+//   res.render('index');
+// });
```
tawebbcn: Remember to remove the code that you don't need so it's more clear and readable

starter_code/views/beers.ejs
```diff
@@ -0,0 +1,5 @@
+<div class="beer-list-container">
+  <% beers.forEach(function(beer){ %>
+  <%- include('partial', {beer: beer}); %>
```
tawebbcn: Try to indent also with ejs

PR #475, opened 2019-01-31
BCN - DAVID REDONDO - WDFT19

rdnd0:


app.js
```diff
@@ -0,0 +1,47 @@
+
+const express = require('express');
+const app     = express();
```
axelgar: Try to put all the requires at the top and separate them form the res. It's cleaner.

PR #479, opened 2019-01-31
BCN-ISMAEL JAOUHAR-WDFS19

mavericucu:


app.js
```diff
@@ -0,0 +1,36 @@
+
+const express = require('express');
+const app     = express();
```
axelgar: Try to put all the requires at the top, separated from the rest. It's cleaner.

PR #481, opened 2019-01-31
Paris Estelle & Delphine

Deie:
@olawinski-e

### DIFF
--- a/starter_code/views/index.hbs
+++ b/starter_code/views/index.hbs
@@ -1,3 +1,9 @@
 <div class="index-container">
   
+    <div class="beer-homepage"></div>
+    <div class="btn-home">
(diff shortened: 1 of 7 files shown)


starter_code/views/index.hbs
```diff
@@ -1,3 +1,9 @@
 <div class="index-container">
   
+    <div class="beer-homepage"></div>
+    <div class="btn-home">
+        <a class="button" href="/beers">Check the Beers!</a>
```
ta-web-paris: another way to target the a tags would be to use the selector in your css `.btn-home a {...}` rather than adding the class name button

ta-web-paris:
What you've done is good! Everything that you started is working and the stying doesn't look bad 💪

PR #484, opened 2019-01-31
PARIS - Devsquad 107 Christina & Jules B.

julesbadin:
done with @CIBailey

### DIFF
--- a/starter_code/views/beers.hbs
+++ b/starter_code/views/beers.hbs
@@ -1,0 +1,6 @@
+<h2>Beers List</h2>
+{{!-- {{#each beersList}}
+<p>{{this.name}}</p>
+{{/each}} --}}
+
+{{> beerPartial}}
--- a/starter_code/views/index.hbs
+++ b/starter_code/views/index.hbs
@@ -1,3 +1,14 @@
 <div class="index-container">
-  
+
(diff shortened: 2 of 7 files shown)


starter_code/views/beers.hbs
```diff
@@ -0,0 +1,6 @@
+<h2>Beers List</h2>
+{{!-- {{#each beersList}}
+<p>{{this.name}}</p>
+{{/each}} --}}
+
+{{> beerPartial}}
```
ta-web-paris: You didn't use the partial exactly how you're meant to. You should loop over your array here and the partial should set the structure for each item - exactly like you did in the version you commented out

starter_code/views/index.hbs
```diff
@@ -1,3 +1,14 @@
 <div class="index-container">
-  
+
+  <div class="beer-image">
+
+    <img src="/images/beer.png" alt="">
+
+    <a class="btn" href="/beers">Check the beers!</a>
```
ta-web-paris: you could have used a more specific tag here to get the links `.beer-image a {...} ` rather than adding an extra class

ta-web-paris:
What you've done is good! You seem to have understood what we covered in class and I like your styling 🙌

PR #490, opened 2019-02-01
Paris - Francesca & Aurelie

AurelieBeno:
@wowfranzik

### DIFF
--- a/starter_code/views/beers.hbs
+++ b/starter_code/views/beers.hbs
@@ -1,0 +1,7 @@
+<ul id="beer">
+  {{#each beers}}
+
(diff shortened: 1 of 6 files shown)


ta-web-paris:
You didn't change the app.js. In your app.js you were meant to have created the routes that link to the beer/random beer pages. When I run the server, if I click the link it give me the `Cannot GET /beers` meaning the route doesn't exist. 

If you'd like, I (Abi) can spend some time reviewing it with you during a break/lunch/after class. The rest of the code looks good, and like it would work but I can't test without the routes

PR #506, opened 2019-02-11
Liran Gibor / Guillermo Aiken - FTWD - MIA

aikengx:


ta-web-mia:
Nice work, this is extremely close to completion. Aside from styling, the only other thing you should do is to pull image_url and other relevant values from the beer object, and have it appear on the page along with the name of the brew. 

One other very small thing that can be fixed is line 4 on the index.hbs. The link for the random beers button had a very small typo on it that broke the link; change it to the following and it should all be good:

`<button><a href="/random-beers">Check a Random Beer!</a></button>
`

Keep up the great work!

PR #507, opened 2019-02-11
Manuel Garaboa & Jack Fischer FTWD-MIAMI 

mannygaraboa:
Ironbeer lab progress so far


ta-web-mia:
So far so good! You can finish off the beers page by pulling the description value from the beer object (the same way you did with the image), and doing some styling to organize it a bit (I recommend flexbox properties). As for the random beer page, try and work on properly directing the page before you worry about any of the data that displays. Keep up the good work 👍

PR #508, opened 2019-02-12
Henry Doce, Nunnapat, Matt Keil - FTWD - Miami

hendo12:


ta-web-mia:
Nice job! The only thing left to do is to pull more values for the random beers page, and to add button  functionality on the homepage. Beers page is responsive too, nice use of flex properties! Keep up the good work you guys 👍

PR #510, opened 2019-02-12
Ric Taglietti FTWD Miami

rtagl:
done, added button functionality


ta-web-mia:
Beers page is grabbing all the necessary parts of the object and the random beer page is working as intended. Very well done!

PR #512, opened 2019-02-12
Victoria Hickam FTWD MIAMI

vhickam:


ta-web-mia:
Great job! The only thing I can think of improving on is styling. Using flex properties for the beers page would make it more responsive, and you could benefit from adding a little bit padding or repositioning the image_url on the random beers page. At that point I'd just be nitpicking though. Keep up the good work 👍

PR #536, opened 2019-03-19
[AMS] Camila & Isabela

camilav2:


starter_code/views/partials/partials.hbs
```diff
@@ -0,0 +1,23 @@
...
+      <img class="beer-img-top" src="{{this.image_url}}" alt="beer-image">
+      <div class="details">
+        <h2 class="beer-title">{{this.name}}</h2>
+        <p class="beer-text">{{this.tagline}}</p>
+        <p class="beer-description">{{this.description}}</p>
+        <p class="beer-food-pairing">{{this.food_pairing}}</p>
+        <p class="brewer-tips">{{this.brewers_tips}}</p>
+      </div>
+</div>
+  {{/each}}
```
lienekechee: Since they have different content, I would create a separate partial for each different kind of card and include the `{{random-beer-partial}}` in the random beer view.  But this works too!

starter_code/views/beers.hbs
```diff
@@ -0,0 +1,17 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <link rel="stylesheet" href="stylesheets/styles.css">
+    <title>Beers</title>
+</head>
+<body>
+    <div class="beers">
+
+      {{> partials}}
+
```
lienekechee: Watch your indentation and spacing :)

starter_code/app.js
```diff
@@ -10,12 +10,33 @@ app.set('view engine', 'hbs');
...
+  .catch(error => {
+    console.log(error)
+  })
+});
+
+app.get('/random-beers', (req, res, next) => {
+  punkAPI.getRandom()
+  .then(beers => {
+    res.render('random-beers',  {y:beers});
+  })
```
lienekechee: Instead of `x` and `y` perhaps name these variables `allBeers` and `randomBeers` so its clear why different partials are used for different data.

PR #569, opened 2019-04-09
PARIS Finished work MIMI ft Duc

ducky92fr:


ta-web-paris:
It looks good and everything works - your css is great 👍

PR #573, opened 2019-04-10
Chris & Thanh PARIS 

ThanhIH:
not finished yet


ta-web-paris:
It seems like you got a little confused during the exercise but you have the right idea. It's cool how you're trying to organise your code with partials, but the set up is not quiet right.

I'm (Abi) happy to spend some time reviewing this lab, or another lab with you this afternoon

PR #610, opened 2019-04-23
MIA FTWD Samuel Yamamoto

YamamotoSamuel:


ta-web-mia:
Well done! If you have extra time, try adding keys from the beer object like the pairing recommendations to the random beers view and to clean up the styling on the random beers page as well. No real complaints on my part though, keep up the good work 👍

PR #611, opened 2019-04-23
MIA-FTWD-APR-Justin Velez

jvelez523:
Worked with Kenneth


ta-web-mia:
Well done! The only thing that can really be improved is some minor styling on the Random Beers page. Nothing to get too hung up on though. Stay focused on what's ahead, don't let HBS being worse than React ruin your day, and keep up the good work 👍

PR #612, opened 2019-04-23
{MIA FTWD} [Jon R.] 

jonriquer:


ta-web-mia:
Everything was pulled from the API properly, and the CSS animations are really nice! Great job, and keep up the good work 👍

PR #620, opened 2019-04-24
MIA-FTWD-Allison

allisonkopp:


ta-web-mia:
Your routing and API calls are all working great, well done! The only thing missing are the article tags in the buttons, since the href tag on its own won't direct to other pages. But it's working in the navbar, so it's nothing to get hung up on. Keep up the good work 👍

PR #670, opened 2019-07-03
MIA-FTWD-Sergio and Billy

sergiovelo17:


ta-web-mia:
Looks good! The only comment I would add is that your beer images change with the window. So they came out warped when I wasn't at full screen.

PR #673, opened 2019-07-03
MIA FTWD - Billy Wong

billywonka3:


ta-web-mia:
You're almost there! Your code looks good and everything is appearing on the console, but I'm receiving an error related to the beer partial that's preventing the views from rendering. Take a look at that when you have the opportunity.
- Nick

PR #675, opened 2019-07-03
MIA-FTWD-Mitch

mitchmalinin:


MarcosPalaciosIII:
Great job on the ironbeers lab. When working with api's, there are times that the documentation is not so straight forward or easy. Getting more practice with api's will give you more experience working with api's which you may find yourself doing a lot of later in your dev career.

Keep up the great work.

PR #794, opened 2019-09-10
MIA FTWD JORGE

jorgeluis42:


ta-web-mia:
Everything's getting pulled up! The only suggestion I would make is maybe make the pictures of the beer a little bigger, but CSS isn't the point of this lab, so don't worry about fixing it, just keep that in mind in the future. Great job, though!

PR #795, opened 2019-09-10
MIA-FTWD-BobbyWang

bobbypwang:
Completed all iterations including the bonus.


ta-web-mia:
Everything is working properly, nice job! The only thing I noticed is the final beer is getting cut off by the footer at the end. Everything else is great.

PR #796, opened 2019-09-10
MIA-FTWD-Enrique Narciso

e-narciso:
think this is finished


ta-web-mia:
Everything checks out. I appreciate you making all the beers fit equally in a row on the display page, it's a nice touch.

PR #805, opened 2019-09-11
MIA - FTWD Osheit Chang

osheitchang:


ta-web-mia:
Now that we're onto the second Ironbeers, looking back on this one, it seems like you have a grasp on this. You're almost home free, good luck on React Ironbeers!

PR #885, opened 2019-11-11
[PAR WDFT] Hanna

hanna974:


ta-web-paris:
Good job! Just watch out it redirects to /random-beers, but I checked and /random-beer works :)

PR #905, opened 2019-11-13
[MAD DE] Rebecca

Rebeccawood:
pictures wont show...  ?
not sure about the random beers.

will need to review tomorrow with clear head.


ta-web-mad:
Pictures are showing, they are even too big for the screen haha

There was no need for a randomBeer partial, since it's only used once in the whole application. Partials are used to avoid repeating code.

Great job!
Teo :monkey: :monkey:

PR #949, opened 2019-12-11
heavily refactor readme

sandrabosk:
- add more details and guidelines in iterations
- have bonus iterations
- move partials as bonus
@josecarneiro Please take a look when you find a moment. Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,15 @@
 ![logo_ironhack_blue 7](https://user-images.githubusercontent.com/23629340/40541063-a07a0a8a-601a-11e8-91b5-2f13e4e6b441.png)
 
-# IronBeers
+# LAB | Express IronBeers
 
 ## Introduction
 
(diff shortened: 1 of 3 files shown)


josecarneiro:
Hi @sandrabosk,

I'm sorry for the huge delay in this. I wanted to go through it mindfully.

Thank you so much for the thorough refactor, this lab was really needing it.

I made a few additions to the refactor. I don't know if you're in agreement with them, we can revert any if you think they're unnecessary.

- I added clarification regarding the inner-workings of the PunkAPI package, without going in depth into specifics.
- I added an extra level of indentation to the iteration titles, since they're under the "Instructions" section and had the same indentation.
- I changed the naming pattern of `hbs` partials to address the issues that we've discussed in previous webdev-collab calls: avoiding uppercase characters in filenames, since different environments resolve file paths differently, and a codebase might run on a students' machine but break in production, while maintaining a valid naming scheme.

Let me know if there's anything else we should look at in this lab.

Thank you!

PR #991, opened 2020-01-30
[BER-WDFT-012020] Elsa & Shahinya

sha-hinya:


starter-code/views/random.hbs
```diff
@@ -0,0 +1,9 @@
+{{> beerpartial this}}
+
+<ul>
+    {{#each this.food_pairing}} <li>{{this}}</li>{{/each}}
```
fenrew: Just so you know: this.food_pairing === food_pairing

PR #994, opened 2020-01-30
[BER-FT] Vicky and Tim

timbotimber:
Pretty happy with this, styling is soooo hard. Maybe we could have a css refresher some time? 

We got everything working ok but we couldn't target the div that contained the beers to make them flex in the style the of the example. Would love to know how to do that!

### DIFF
--- a/starter-code/app.js
+++ b/starter-code/app.js
@@ -13,7 +13,31 @@ app.use(express.static(path.join(__dirname, 'public')));
 
 // add the partials here:
 
+hbs.registerPartials(path.join(__dirname, 'views/partials'));
+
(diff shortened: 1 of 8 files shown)


fenrew:
Nice website!:) I will be more than happy to give some css advice. You should come over when you have the time today and I can show you how I went by to find the correct class/div for your main beer page (if I have understood your issue that is...).
You have given that container the class .homepage already

timbotimber:
Finally read this! Let's talk about it on Monday if we have time. Maybe we could have a CSS refresher after class one day next week? Would be good to recap it.

PR #998, opened 2020-01-30
[BERFT0120] Shaeka-Vaso

vasilikibalidou:


vasilikibalidou:
We couldn´t figure out iteration 6, so we commented out the part that we wrote because it wasn´t working properly.

starter-code/app.js
```diff
@@ -12,8 +12,39 @@ app.set('views', path.join(__dirname, 'views'));
...
+    .getRandom()
+    .then(responseFromAPI =>
+      // your magic happens here
+      //console.log('Beers from the database: ', responseFromAPI)
+      res.render('random-beers', { beer: responseFromAPI })
+    )
+    .catch(error => console.log(error));
+});
+
+/* app.get('/beers/beer-:id', (req, res) => {
```
DalinaLeila: `app.get('/beers/beer-:id', (req, res) => {`

DalinaLeila: `const id = req.params.id`

PR #999, opened 2020-01-30
FT Berlin Lluis & Laurel - done

LluisManas:


starter-code/views/random-beers.hbs
```diff
@@ -0,0 +1,17 @@
+<div>
+    {{#each randomBeers}}
+    <div>
+
+        <div class="beer-image">
```
DalinaLeila: this is the same as your beer partial, so you can just use that instead of rewriting code :)

PR #1002, opened 2020-01-30
BE_FT_012020_Mohammed_Akua

akups:


starter-code/views/index.hbs
```diff
@@ -1,3 +1,6 @@
+<h1>Ironbeers</h1>
 <div>
-
+    <img src="images/beer.png" alt="beers">
+    <button onclick="/beers" type="">Check the beers</button>
```
fonsogms: here to redirect to a new route you should have tried this:
```
<a href="/route">
<button type="">Check the beers</button>
</a>


```
Putting the button inside of a `<a>` tag

PR #1004, opened 2020-02-03
[PAR - WDFT 0120] Anthony B. - ongoing

abrebion:


Jepassaispar:
Great job as usual, again good use of await, just for your route get /beers you did the old way of the non-cool kids which is good as well but you are defining your const beers like in your await calls and that's not useful.

I said that because I really have nothing else to say so again great job :)

PR #1007, opened 2020-02-03
[PAR-WD-FT] - Nina - ongoing

Gorlanova:


Jepassaispar:
Great job :)
Only thing is that I think it can get a bit confusing in your routes : 
```
app.get('/beers', (req, res) => {
  punkAPI
    .getBeers()
    .then(beers => {
      res.render('beers', { array: beers });
    })
    .catch(err => console.error(err));
});

app.get('/random', (req, res) => {
  punkAPI
    .getRandom()
    .then(beers => {
      res.render('random', { beers });
    })
});
```
For the first route I would have called your `{array: beers}` directly `{beers}` like what you did for the random route, so on your beers.hbs you would be iterating on `{{#each beers}}` instead of `{{#each array}} ` (both are working, it's just the syntax).
For your second route, you are receiving from your api a random beer object inside an array but you know that every call to the db you will have only one object in this array so instead of sending the array to render with the page and iterating a length 1 array, I would have done like following : 
```
    .then(beers => {
      const randomBeer = beers[0];
      res.render('random', { randomBeer });
    }) 
```
and so you won't have to do a each on your random.hbs ; 
```
<section class="page">
    <div class="random">
        <img class="img" src="{{image_url}}" alt="{{name}}">
        <h3>{{name}}</h3>
        <h4>{{tagline}}</h4>
        <p>{{description}}</p>
        <ul>
            {{#each food_pairing}}
            <li>{{this}}</li>
            {{/each}}
        </ul>
        <p>{{brewers_tips}}</p>
    </div>
</section>
```

Except that it's all good and I like the fact that you're revisiting the style :) (maybe navbar is a bit big... But hey, who cares !)

Gorlanova:
Haha I like huge navbars these days… 
Thank you for the feedback ! I struggled with the array of one, and then got a bit lost in the process but it’s very clear now  :)

PR #1015, opened 2020-02-03
[PAR - WDFT] - Coline - ongoing

colinevel:
I tried a bunch of stuff, but I didn't manage to do the bonus Iteration 7.
It doesn't seem to recognize the ID when I click on the name of a beer..


Jepassaispar:
Viens me voir pour qu'on voit ensemble dès que tu as un moment :)

PR #1021, opened 2020-02-04
[PAR][012020][WDFT][LABIRONBEERS]TAYSSIR ongoing

taytay81:
I could not manage to put some style on the random beer as they share the same partial just wanted to know how to do that .
thank you


clarabesset:
Great job, the CSS is more than good enough, nice !!

PR #1023, opened 2020-02-05
[PAR - WDFT] Remy Pham ongoing

kurosaek:


clarabesset:
Nice job! You made the CSS work for all the beers, even when they don't all have any description, great work!

PR #1068, opened 2020-02-12
[LIS JAN 20 - Manuel Rocha]

TemidoRocha:


TemidoRocha:
@ironhackLis-webTeam

i tried to make the search.

i inserted the action on the form
the name on the input

I think it is on the app.js side

My idea is to seach a id and redirect for the singlebeer page

PR #1117, opened 2020-03-11
AMS-WEB DEV FT-FEB2020- Francis Bukenya

Frankie-B:
This is only the mandatory iterations. I will continue to work on this to finish bonus iterations as well.


starter-code/public/stylesheets/footer.css
```diff
@@ -0,0 +1,15 @@
+.footer-container {
+  display: flex;
```
mswidzinska: cq--
You should use not a display:flex, but fixed position, here:
position: fixed;
bottom: 0;  
right: 0;

Frankie-B:
@mswidzinska I do not think that it allows me to make a second pull request since this one is open. But I made the changes that you suggested to me.

PR #1210, opened 2020-04-20
PAR WDFT Sam L ongoing

samlfair:


TatijanaRajcic:
Hey Sam! Your lab is perfect!
- I don't have anything to say! I just have 1 quick question:  what's the use of "extended=true" that you are passing to your partials? I removed it, and it does not change anything
- nice use of clickable=true , it's the first time I see this helper :)

Well done Sam!

samlfair:
Thanks! I try to use the `extended` variable with an `if` handlebars block to show more and less information on the "beers" page versus the "one-beer" page. I might not have implemented it right though 😅

PR #1219, opened 2020-04-21
[PAR-WDFT-0320]Olivier Brulé

OBxel:
without bonus


ta-web-paris:
Good job Olivier !
You'll have time to practice CSS during project 2 ;) 

 Nina

PR #1254, opened 2020-05-27
[REMOTE-WDFT-0520] MICHELLE LOCK

michelleytlock:
Finished all the mandatory parts, will work on bonus tomorrow morning


michelleytlock:
Couldn't get partial to work, but updated with styles.

PR #1360, opened 2020-07-27
[WDPT - Remote - 062020] - Alejandro Gonzalez

alejanglez:
All Iterations completed + Bonus!

// Few check of App.js if everything look fine. I saw sometime in ramdom beer they use--- responseFromAPI[0]
But it isnt working for me//

Thank you !!

### DIFF
--- a/app.js
+++ b/app.js
@@ -13,13 +13,32 @@ app.set('views', path.join(__dirname, 'views'));
 app.use(express.static(path.join(__dirname, 'public')));
 
 // Register the location for handlebars partials here:
-
+hbs.registerPartials(path.join(__dirname, 'views/partials'));
 // ...
 
 // Add the route handlers here:
 
 app.get('/', (req, res) => {
   res.render('index');
+});
+
+app.get('/beers', (req, res) => {
+  punkAPI
+    .getBeers()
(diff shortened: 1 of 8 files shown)


app.js
```diff
@@ -22,4 +22,23 @@ app.get('/', (req, res) => {
...
+      console.log(beersFromApi);
+    })
+    .catch(error => console.log(error));
+});
+
+app.get('/random-beer', (req, res, next) => {
+  punkAPI
+    .getRandom()
+    .then(responseFromAPI => {
+      res.render('random-beer', { responseFromAPI });
```
valmgisbert, line 39: Yes! what you say in your comment is another way of getting the data, in which you extract the random beer from the array and you send it as an object. Though if you want to use partials with this it might give you issues if you go with this approach.

views/random-beer.hbs
```diff
@@ -0,0 +1,9 @@
+{{#each responseFromAPI}}
+<div>
+  {{> beerPartial this}}
```
valmgisbert, line 3: nice use of partials!

valmgisbert:
Good job Ale! Whenever you can, work on the final bonus on how to get an individual beer with their id.

PR #1364, opened 2020-07-27
[REMOTE-WDPT-0620] Debora Crescenzo

NubivagantYansa:
Done plus bonuses.
The styling of 'Random beer' page is ugly ha!
I managed to complete successfully bonus 6 looking up the documentation, but I think I would benefit if someone could talk me through it to consolidate the knowledge.
Thanks!

### DIFF
--- a/app.js
+++ b/app.js
@@ -14,12 +14,45 @@ app.use(express.static(path.join(__dirname, 'public')));
 
 // Register the location for handlebars partials here:
 
-// ...
+hbs.registerPartials(path.join(__dirname, 'views/partials'));
 
 // Add the route handlers here:
 
 app.get('/', (req, res) => {
   res.render('index');
+});
+
+//gets list of beers
+app.get('/beers', (req, res) => {
+  punkAPI
+    .getBeers()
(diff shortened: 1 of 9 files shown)


app.js
```diff
@@ -14,12 +14,45 @@ app.use(express.static(path.join(__dirname, 'public')));
...
+    .catch(error => console.log(error));
+});
+
+//gets one random beer
+app.get('/random-beer', (req, res) => {
+  punkAPI
+    .getRandom()
+    .then(responseFromAPI => {
+      res.render('random-beer', { responseFromAPI });
+      console.log(responseFromAPI);
```
valmgisbert, line 41: after checking your result in the terminal, remember to clean your console.logs!

valmgisbert:
Great job Debora! you did well in your approach. Basically what you do in the bonus 6 is to look for the specific beer by one of its attributes; its' id in this case (since each id is unique, it is easier to filter through tons of data and be more specific about the search results). You make a request to the API giving it that parameter, and so it pulls the object that matches that parameter. If you need to talk a bit more about the bonus, please let me know!

PR #1365, opened 2020-07-28
[REMOTE-WDPT-0602] -Cristina Matamoro

cris-developer:
Done till iteration 6  and added some bootstrapp style.
I will try to finish today iteration 6 and add more styling (iteration 7). (Styling is always more time consuming than i  initially thought)


app.js
```diff
@@ -14,10 +15,36 @@ app.use(express.static(path.join(__dirname, 'public')));
...
+  res.render('index.hbs');
+});
+
+//Iteration 3.1_route endpoint beers
+
+app.get('/beers', function (req, res) {
+  punkAPI
+    .getBeers()
+    .then(beersFromApi => {
+      console.log(beersFromApi);
```
valmgisbert, line 32: after you see the results you want in your terminal, remember to clean your console.logs!

app.js
```diff
@@ -3,6 +3,7 @@ const express = require('express');
 const hbs = require('hbs');
 const path = require('path');
 const PunkAPIWrapper = require('punkapi-javascript-wrapper');
+const { response } = require('express');
```
valmgisbert, line 6: this const is used to load the express library, and doesn't need to be deconstructed. Is there a reason why you used it like this? was it required by vscode?

app.js
```diff
@@ -14,10 +15,36 @@ app.use(express.static(path.join(__dirname, 'public')));
 
 // Register the location for handlebars partials here:
 
-// ...
-
 // Add the route handlers here:
+hbs.registerPartials(path.join(__dirname, 'views/partials'));
 
+// Iteration 2_route endpoint home
+app.get('/', (req, res) => {
```
valmgisbert, line 22: is there a reason why you have the index route twice? also, did it not work without adding the .hbs suffix to your renders? it should read it automatically

valmgisbert:
Nice work Cristina! whenever you can, please finish the other bonuses.

PR #1366, opened 2020-07-28
[WDPT REMOTE 0602] Lynn W-B

lynn-w-b:
Iterations 1-4 complete (without styling)

### DIFF
--- a/app.js
+++ b/app.js
@@ -11,12 +11,7 @@ app.set('view engine', 'hbs');
 app.set('views', path.join(__dirname, 'views'));
 
 app.use(express.static(path.join(__dirname, 'public')));
-
-// Register the location for handlebars partials here:
-
-// ...
-
-// Add the route handlers here:
+hbs.registerPartials(path.join(__dirname, 'views/partials'));
 
 app.get('/', (req, res) => {
   res.render('index');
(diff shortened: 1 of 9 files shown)


lynn-w-b:
Got as far as I could with the styling in the time I have available. Can you please give me some feedback re the getting the beer id from the url part? Maybe we could chat this through? I followed the instructions online and it works but I'm not sure I quite grasp why. Thanks

app.js
```diff
@@ -11,15 +11,33 @@ app.set('view engine', 'hbs');
...
+  .catch(error => console.log(error));
+});
+app.get("/random-beers", (req, res, next) => { 
+   punkAPI.getRandom()
+   .then(randomFromApi => { 
+    res.render("random-beers",{randomFromApi});
+   })
+   .catch(error => console.log(error));
+ });
+ app.get("/getbeer/:id", (req, res, next) => {
```
valmgisbert, line 33: good structure in calling the ID!

views/random-beers.hbs
```diff
@@ -0,0 +1,17 @@
+<div id="random">
+    {{#each randomFromApi}}
+        {{> beerpartial this}}
+        <div id="food">
+        <h5>Food Pairing:<br></h5>
+        <ul>
+        {{#each food_pairing}}
+        <li>{{this}}</li>
```
valmgisbert, line 8: work a bit on the spacing here. All is good tho!

valmgisbert:
Great work Lynn! and sure! what you're doing in it 6 is making a request to the API with a certain parameter: in this case their id. By looking through their beer array, that request plucks the info of the one matching the id inside of the target, and sends it to be reflected in your hbs file. If you need any more information about this, please let me know!

PR #1368, opened 2020-07-28
[REMOTE-WDPT-0602] - Matthew Howard

mattalanhoward:
Would like to see how the id is done after adding link for each beer (Iteration 6).

What is the best way to make a footer stick to the bottom?  What I did mostly works but now I have a bit of a scroll on the main page and random beer page that I do not like.

### DIFF
--- a/views/beers.hbs
+++ b/views/beers.hbs
@@ -1,0 +1,10 @@
+<div>
+<section class="beers">
+    <ul>
+        {{#each responseFromAPI}}
+            {{>beerpartial}}
+        {{/each }}
+    </ul>
+</section>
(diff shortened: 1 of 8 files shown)


views/beers.hbs
```diff
@@ -0,0 +1,10 @@
+<div>
+<section class="beers">
+    <ul>
+        {{#each responseFromAPI}}
```
itstheandre, line 4: quick piece of advice: responseFromAPI might not be the best name. Maybe beersList, or another name that you can immediately know what you are looking at if you are looking at code in a week and a half from now

PR #1370, opened 2020-07-28
[REMOTE-WDPT-0602] - Giovanna Izzo

Hachiko9:
did not manage to do the style

### DIFF
--- a/views/beer-detail.hbs
+++ b/views/beer-detail.hbs
@@ -1,0 +1,5 @@
+<div>
+    {{#each beer}}
+        {{> beer}}
+    {{/each}}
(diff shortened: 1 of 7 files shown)


views/beer-detail.hbs
```diff
@@ -0,0 +1,5 @@
+<div>
+    {{#each beer}}
+        {{> beer}}
```
itstheandre, line 3: in this situation, I would feel more comfortable if you names these somehow different things. I know what I am reading, but this can get confusing super fast.
Always code assuming that the next time you see this you have no idea what you are looking at :)

PR #1373, opened 2020-07-30
[WDFT-REM-2007] ALVARO, BODEI AND LANETTE

AlvaroSLam:


app.js
```diff
@@ -12,14 +12,45 @@ app.set('views', path.join(__dirname, 'views'));
...
+
 // Register the location for handlebars partials here:
 
 // ...
 
 // Add the route handlers here:
 
 app.get('/', (req, res) => {
-  res.render('index');
+  res.render(__dirname + '/views/index');
```
IH-WebDev-TA-Remote, line 24: since there is a code line in L11 to set the views in the __dirname + / views you don't need to redeclare the path here, just by putting index or index.hbs it should locate and render the correct hbs.

app.js
```diff
@@ -12,14 +12,45 @@ app.set('views', path.join(__dirname, 'views'));
...
+  .catch(error => console.log(error));
+});
+
+app.get('/beers/:beerId', (req, res) =>{
+
+  let beerId = req.params.beerId;
+  punkAPI.getBeer(beerId)
+    .then((data)=> {
+      let myBeer = data[0]
+      console.log(data[0].image_url)
```
IH-WebDev-TA-Remote, line 51: remember to remove lines of code that were used only to test, since you are passing the whole data element in line 52, then these two lines (50 and 51) are not used/needed.

views/partials/beerpartial.hbs
```diff
@@ -0,0 +1,2 @@
+
+<div>{{this.name}} <img src="{{this.image_url}}" alt="img"> {{this.description}} {{this.tagline}}</div>
```
IH-WebDev-TA-Remote, line 2: As best practices, write your div content in separate lines for readability. More important, never write your html without html tags (see description and tagline)

IH-WebDev-TA-Remote:
Good job, all seems to be working good. Styles were not important for this lab but that doesn't mean you don't need to write a good html code. Please remember to use the correct tags and never put html elements without them.
Good use of partials and the links for each beer. Keep up the good work! :)

PR #1375, opened 2020-07-30
[WDFT-REMOTE-2007] Gunner, Alejandro

alejandrohj:


views/index.hbs
```diff
@@ -1,3 +1,5 @@
-<div>
-
+<div class="index">
+<img src="/images/beer.png">
```
IH-WebDev-TA-Remote, line 2: watch indentation of html elements

IH-WebDev-TA-Remote:
Very good job! and even some styles, kudos. Try to check the solution on how the links to each beer are done via params, this is very important as it is something used often.

PR #1376, opened 2020-07-30
[web remote 2007] Gonzalo, Rafael

gonzalo-development:


views/index.hbs
```diff
@@ -1,3 +1,8 @@
 <div>
+<img src="/images/beer.png" alt="beer">
```
IH-WebDev-TA-Remote, line 2: watch indentation of elements

IH-WebDev-TA-Remote:
Good job! Only important thing missing was using params to create links for each beer, take a look in the solutions at how this is done, as it is something that is used very often.

PR #1378, opened 2020-07-31
[WDFT-REM-2007] Iron-Beer LAB / Elise and Joel

JoelRocaMartinez:


views/partials/beerpartial.hbs
```diff
@@ -0,0 +1,7 @@
+<div>
+<img src={{this.image_url}} alt="beer">
```
IH-WebDev-TA-Remote, line 2: watch indentation of html elements

IH-WebDev-TA-Remote:
Very good job! and even throwing some styles. Kudos :) Only important thing missing was using params to create links for each beer, take a look in the solutions at how this is done, as it is something that is used very often.

PR #1394, opened 2020-08-31
[WDPT-AMS20-Yuka]

yukajyeh:


views/beers.hbs
```diff
@@ -0,0 +1,11 @@
+<ul>
```
guido-an, line 1: If you want to display the beers on 3 columns you can add a class here and then add  `
`.your-class {`
   ` display: flex;`
   ` flex-wrap: wrap;`
}`

Then give _width: 33%_ to your .beers class.

This should be the final result: https://prnt.sc/ua6r9o

PR #1440, opened 2020-09-18
[PAR-WDFT-248] FRANCOIS VIEUX 

Georges-Seonho:


ta-web-paris:
Nice job ! No CSS :(( but it was bonus so all good

Nina

PR #1441, opened 2020-09-18
[PAR-WDFT-248] Jorell Langmar

JorellLangmar:


ta-web-paris:
Nice job ! The look is hardcore but it was a bonus so all good :)

Nina

PR #1442, opened 2020-09-18
[PAR-WDFT-248] Bey Bouziane

Bouziane95:


ta-web-paris:
Nice job !! Everything is working well :)


Nina

PR #1444, opened 2020-09-18
[PAR-WDFT-248] Luca Marongiu

AyloSrd:
[PAR-WDFT-248] Luca Marongiu

### DIFF
--- a/app.js
+++ b/app.js
@@ -30,4 +30,12 @@ app.get('/beers', async (req, res) => {
   }
 })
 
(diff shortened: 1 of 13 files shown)


ta-web-paris:
Hey Luca !
Nice job and nice CSS !
Regarding the image : you can set "margin: auto" on the img and it will center them :)
(margin auto only sets margins for left and right, and divides the remaining space so it's centered)

Nina

PR #1447, opened 2020-09-20
[PAR WDFT]Laurent Vuillaume

figura01:


ta-web-paris:
Very nice Laurent ! Great CSS
The only thing : if I understood correctly, you called your css in the server (when rendering) ? If so, remember next module or when working in groups to avoid that and do all front job in the front part because it can be confusing when trying to locate an error ! People will often work on front OR back, and will not check other files than their zone :)

Otherwise great job

Nina

PR #1448, opened 2020-09-20
[PAR-WDF-248]Tatiana Matzenbacher

tatimatz:


ta-web-paris:
Hey Tatiana ! 
Nice start, a few points to help :

- Your homepage image is not displaying on the homepage because in hbs (and later, React) you don't need to say "public" (for img, audio, etc etc... the browser knows that you'll take the media files from the public folder, because of app.js line 14) so if you remove "public/" from the path, it works !

- in the beers page, images are not displayed because you made a mistake : you should create an "img" tag with the image url as a "src" attribute like this : 
``
and be careful : it's "this.image_url" and not "this.img" ! You can check that when you console log the response from the api (you have every key listed)

- when checking for a random beer it's not working : you have an error in the terminal console (app.js line 33 : you console log something that doesn't exist, you should console log "apiRandom" instead because that's the name of the variable (line 32) where you stored the response from the api call for the random beer)

- when you correct that, we can access the random beer page but nothing is displayed : it's because you don't render the right elements : when you check your random-beer.hbs file, you have two mistakes : the image thing (as earlier) and the way you call your elements.

In the "beers" view, you called "this.name" etc etc because we already knew what "this" is : you say {{#each beers}} before. That means that the backend is sending a list of information (stored in a variable called "beers", as you said in the file app.js line 25 and later sending it to the frontend in line 28 with the variable beers that contains the api response !)

It should be the same logic for the random-beer, except that you don't have a loop in your hbs file, since it's only one beer : so you don't have {{#each randomBeer}} so your browser doesn't know what "this" refers to. we need to define exactly so it can understand it has to get the info from the variable randomBeer (as you said line 34 : the page random-beers receives the apiRandom[0] that is renamed "randomBeer") so we need to write our info like this :
`this.randomBeer.xxx`

and voilà it works !

Tell me if my details are not clear, I can try to explain better !
Make these changes and push again so we can validate your lab ;)

Nina

PR #1451, opened 2020-09-20
[PAR - WDFT - 248] Timothée NICOLE

timothee-nicole:


Fanny-Sbn:
Hello Timothée, thanks for your lab. 

Nice job for the mandatory iterations and the first bonus. 

Bonus 6 is not working, you didn't put the same href than your route in your displayBeer partial. (It's just a small mistake, so try to correct it by yourself ;) )
The CSS isn't done so it's ugly hahaha but it was a bonus so it's ok. If you have time, try to do it ! 

Good job Timothée ! 

Fanny

PR #1452, opened 2020-09-20
[PAR-WDFT-248]Marina Fuxa

KiwiMantelli:
iteration 1 done
iteration 2 done
iteration 3 done
iteration 4 it doesn't work
iteration bonus 5 not yet
iteration bonus 6 not yet
iteration bonus 7 done


Fanny-Sbn:
Hey Marina !

Nice start, a few points to help :

-Have you tried to check the result in the browser ? Actually you can't and you have an error in your console. You tried to get a router with a path that is outside your folder app... what did you try to do ? (you can remove this line !!!)

After removing the previous line, it's working but : 

-Your homepage image is not displaying on the homepage because in hbs (and later, React) you don't need to say "public" (for img, audio, etc etc... the browser knows that you'll take the media files from the public folder, because of app.js line 14) so if you remove "public/" from the path, it works !

-When checking for a random beer it's working but it's not optimized. Since it's only one beer, you could send as a view parameter the only beer that you need and then call it in your randomBeer view.  

app.get('/randomBeers', async (req, res) => {
  const rand = await punkAPI.getRandom();
  try {
    res.render('randomBeers', { oneBeer :rand[0] });
    console.log(rand);
  } catch (err) {
    console.error(err);
  }
});

Try to change your view in consequence so that it can work without a each. 

By the way, the name of your route is not the one they ask you, it's supposed to be /random-beers. 

The CSS is not really good tough... But it was a bonus so it's ok. 

Make these changes and push again so we can validate your lab ;)

Fanny

PR #1453, opened 2020-09-20
[PAR WDFT 248] Fabien Marchand

fabienMarchand:


Fanny-Sbn:
Hello Fabien, thanks for your lab. 

Good job, everything is working well ! The CSS could be better but it was a bonus so it's ok.

Fanny

PR #1455, opened 2020-09-20
[PAR-WDFT-248] Nicolas Enault

nenault:


Fanny-Sbn:
Hello Nicolas, thanks for your lab. 

Very good job, everything is working well and nice use of the router ! The only thing I can say is that the CSS could be better, look at the buttons and the nav bar, they have box-shadows etc,... Anyway it was a bonus so it's totally ok ! 

 Good job ! 💯

PR #1457, opened 2020-09-22
WDFT-PARIS-AUG20-LAURA MANTAL-D'ABOVILLE

Lauradabo:


Fanny-Sbn:
Hello Laura, thanks for your lab ! 

Really good job ! 

Just one thing : 
When checking for a random beer it's working but it's not optimized. Since it's only one beer, you could send as a view parameter the only beer that you need and then call it in your randomBeer view.

app.get('/random -beer', async (req, res) => {
const randomBeer = await punkAPI.getRandom();
try {
res.render('random-beer', { oneRandomBeer : randomBeer[0]});
} catch (err) {
console.error(err);
}
});

Try to change your view in consequence so that it can work without a each.

Bonuses : You could have done bonus 6 and the CSS of the random-beer view can be improved ! The rest is great :) 

Fanny

PR #1458, opened 2020-09-22
[PAR-WDFT-248] Olivo Raphaëlle

RaphOL:


Fanny-Sbn:
Really good job Raphaëlle, it's working well ! 

I would love to have a finished CSS and the other bonuses, but it wasn't mandatory so it's totally ok ! ;)

Fanny

PR #1464, opened 2020-10-08
[REM WDFT 2009]Aleks, Logan, Marta

MartaObregon:


views/random-beer.hbs
```diff
@@ -0,0 +1,11 @@
+
+    {{#each responseFromApi}}
+    <div>
+        <img src="{{this.image_url}}">
+        {{this.name}}
+        {{this.description}}
+        {{this.tagline}}
+        {{this.food_pairing}}
```
IH-WebDev-TA-Remote, line 8: since foor_pairing is another array you could have use another # each to list the elements.

IH-WebDev-TA-Remote:
Good job! all seems to work correctly and the code is clean and easy to read. Although don't be afraid to try those bonuses! Working with partials could really reduce the amount of code in your projects. Keep it up!

PR #1495, opened 2020-11-03
[FTAMS102020] Rich Alossi

alossix:


app.js
```diff
@@ -13,13 +14,34 @@ app.set('views', path.join(__dirname, 'views'));
...
 app.get('/', (req, res) => {
-  res.render('index');
+  res.render('index.hbs');
+});
+
+app.get('/beers', (req, res) => {
+  punkAPI
+    .getBeers()
+    .then(beersFromApi => {
+      res.render('beers.hbs', { beersFromApi });
```
LloydC, line 29: Please remove your `console.log();` after you're finished confirming that everything works.

app.js
```diff
@@ -13,13 +14,34 @@ app.set('views', path.join(__dirname, 'views'));
...
+  punkAPI
+    .getRandom()
+    .then(randomBeer => {
+      res.render('random-beer.hbs', { randomBeer });
+    })
+    .catch(error => console.log(error));
+});
+
+app.listen(PORT, () => {
+  console.log('🏃‍ on port ${PORT}');
```
LloydC, line 46: I think you used single quotes instead of backticks here
```
app.listen(PORT, () => {
  console.log(`🏃‍ on port ${PORT}`);
})

File: app.js [hunk 3]
```diff
@@ -13,13 +14,34 @@ app.set('views', path.join(__dirname, 'views'));
...
+    })
+    .catch(error => console.log(error));
 });
 
-app.listen(3000, () => console.log('🏃‍ on port 3000'));
+app.get('/random-beer', (req, res) => {
+  punkAPI
+    .getRandom()
+    .then(randomBeer => {
+      res.render('random-beer.hbs', { randomBeer });
```
LloydC, line 40: Thanks to line 12 `app.set('view engine', 'hbs');`, Express is able to add the extension `.hbs` by itself when using `res.render()`.

So you don't need to add it, you can refer to the file name only and it'll work the same
```
res.render('random-beer', { randomBeer });

File: views/random-beer.hbs [hunk 4]
```diff
@@ -0,0 +1,23 @@
...
+<html lang="en">
+
+<head>
+    <title>Random Beer</title>
+</head>
+
+<body>
+    <div class="card">
+        {{#each randomBeer}}
+        <div class="imgholder">
```
LloydC, line 11: You can replace line 11-16 with your `beerPartial.hbs`
```
{{>beerPartial this}}

PR #1497, opened 2020-11-03
[WD-FT-AMS-102020] Yann Quintin

YannQuintin:
Here are the action taken so far:
1. Iteration 1, 2 and 3 are covered.
2. Iteration 3.1 is unclear to me. Replicating the approach used for the NBA players, I've added folders and corresponding files including controllers, routes and views.
3. I've called the .getBeers() method in the beers.controllers.js file. Which seemed to be the best place at that point. 

Current issues:
- When running 'node app.js' I see 'Error: Cannot find module '../controllers/beers.controller' ' which prevents me from see my implementation.
- Uncertainty around the folder structure and if all these routes & controllers are necessary, or if a simpler file/folder architecture is required.

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,12 +1,12 @@
 const express = require('express');
-
+const port = 3000;
 const hbs = require('hbs');
 const path = require('path');
 const PunkAPIWrapper = require('punkapi-javascript-wrapper');
 
-const homeRouter = require("./routes/home.routes");
+/* const homeRouter = require("./routes/home.routes");
 const beersRouter = require("./routes/beers.routes");
-const randomBeersRouter = require("./routes/random-beers.routes");
+const randomBeersRouter = require("./routes/randombeers.routes"); */
 
 const app = express();
 const punkAPI = new PunkAPIWrapper();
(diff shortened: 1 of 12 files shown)


app.js
```diff
@@ -13,13 +17,18 @@ app.set('views', path.join(__dirname, 'views'));
...
 
 // Register the location for handlebars partials here:
-
+hbs.registerPartials(__dirname + "/views");
 // ...
 
 // Add the route handlers here:
+app.use("/", homeRouter);
+app.use("/beers", beersRouter);
+app.use("/teams", randomBeersRouter);
```
LloydC, line 26: It should be `app.use("/random-beers", randomBeersRouter);`

app.js
```diff
@@ -13,13 +17,18 @@ app.set('views', path.join(__dirname, 'views'));
...
 
-app.get('/', (req, res) => {
+app.get('/', (req, res) => { //Original setup
```
LloydC, line 28: You have route handlers on line 24-26 so you don't need line 28-30

app.js
```diff
@@ -13,13 +17,18 @@ app.set('views', path.join(__dirname, 'views'));
...
   res.render('index');
 });
 
 app.listen(3000, () => console.log('🏃‍ on port 3000'));
+
+module.exports = router;
```
LloydC, line 34: There's no "router" in app.js, what you are defining in this file is `const app = express();` therefor it should be
```
module.exports = app;

File: controllers/beers.controllers.js [hunk 4]
```diff
@@ -0,0 +1,6 @@
+exports.renderBeers = (req, res) => {
+    const beers = punkAPI.getBeers()
+    .then(beersFromApi => console.log('Beers from the database: ', beersFromApi))
+    .catch(error => console.log(error));
+    res.render("beers", { beers });
```
LloydC, line 5: You can't render the beers here because ` res.render("beers", { beers });` is outside the `.then()` scope where the beers data is available.

The correct way to pass that data is
```
exports.renderBeers = (req, res) => {
    const beers = punkAPI.getBeers()
    .then(beersFromApi => res.render("beers", { beers });)
 
}

File: controllers/randombeers.controllers.js [hunk 5]
```diff
@@ -0,0 +1,3 @@
+exports.renderIndex = (req, res) => {
```
LloydC, line 1: This is literally a copy of `home.controller.js`, please remember to update your code immediately when you're copying code otherwise you'll miss/forget some things.

You also need to create a `random-beers.hbs` file.

routes/randombeers.routes.js
```diff
@@ -0,0 +1,9 @@
+const express = require("express");
+
```
LloydC, line 2: Same comment as above, update your code as soon as you have copied it from somewhere else according to what you're trying to do.

views/index.hbs
```diff
@@ -1,3 +1,19 @@
...
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@4.5.3/dist/css/bootstrap.min.css" integrity="sha384-TX8t27EcRE3e/ihU7zmQxVncDAy5uIKz4rEkgIXeMed4M0jlfIDPvg6uqKI2xXr2" crossorigin="anonymous">
+    <title>Home</title>
+</head>
+<body>
+    <div>
+        <img src="/public/images/beer.png" class="rounded mx-auto d-block" alt="pint">
+        <button href="/beers" type="button" class="btn btn-primary">Check the beers!</button>
```
LloydC, line 12: You can only use the "href" attribute on `<a>` or `<link>`.

The correct way to write line 12 is 
```
<a href="/beers">
  <button  type="button" class="btn btn-primary">Check the beers!</button>
</a>

File: views/index.hbs [hunk 8]
```diff
@@ -1,3 +1,19 @@
...
+        <button href="/random-beers" type="button" class="btn btn-primary">Check a Random Beer!</button>
```
LloydC, line 13: Same as above
```
<a href="/random-beers">
  <button  type="button" class="btn btn-primary">Check a Random Beer!</button>
</a>

PR #1498, opened 2020-11-03
 [FTAMS102020] Sabela Pena Garcia 

defSabe:


app.js
```diff
@@ -1,25 +1,53 @@
...
+      });
+    })
+    .catch(error => console.log(error));
+});
+
+app.get('/random-beers', (req, res) => {
+  punkAPI
+    .getRandom()
+    .then(randomBeer => {
+      res.render('random-beer.hbs', {
```
LloydC, line 42: You're missing an "s" in your file name so the page didn't render.
```
res.render('random-beers.hbs', { randomBeer })

File: views/random-beers.hbs [hunk 2]
```diff
@@ -0,0 +1,10 @@
+<div>
+    {{#each beersFromApi}}
```
LloydC, line 2: The name of the property you're sending from `app.js` is `randomBeer` here so that's the one you should be using
```
{{each randomBeer}}

PR #1500, opened 2020-11-04
[FT-AMS-102020] Elia Chiarucci

eliachiarucci:


app.js
```diff
@@ -22,4 +24,29 @@ app.get('/', (req, res) => {
...
+app.get('/random-beers', (req, res) => {
+  punkAPI.getRandom()
+  .then((beer) => {
+    res.render('random-beer', {beer});
+  })
+  .catch(e => res.status(500).send("Something went wrong"));
+})
+
+
 app.listen(3000, () => console.log('🏃‍ on port 3000'));
```
LloydC, line 52: Just wanted to mention that it's considered common practice to store the port in a variable `const port = 3000;` then you can refer to it `app.listen(port, () => console.log(`🏃‍ on port ${port}`))`

PR #1502, opened 2020-11-04
[FTAMS102020] Nina van den Braak

ninsajala:
Still want to do bonus iteration 6, will look into that later


app.js
```diff
@@ -1,25 +1,43 @@
...
-app.get('/', (req, res) => {
-  res.render('index');
+app.get(`/beers/beer-:id`, function (req, res) {
+  punkAPI.getBeer((req.params).id).then(beerID => res.render(`beer-by-id`, {
+    beerID
+  })).catch(error => error)
 });
 
-app.listen(3000, () => console.log('🏃‍ on port 3000'));
+app.listen(3000);
```
LloydC, line 41: It's considered good practice to have a callback to console.log which port your server is running on, I also recommend you get into the habit of storing your port number in a variable (it's commonly done as well)
```
const PORT = 3000;

app.listen(PORT, () => console.log('Server is running on port ${PORT}'));

ninsajala: Thanks! Will make it a habit

PR #1506, opened 2020-11-04
[FTAMS102020] Emma Northridge

emmanorthridge:
Completed up to iteration 6. 
Functionality working, but haven't finished styling.


wax-bear (commented):
As you already mention in the commit message, more styling could be applied.
In several places I see empty lines which could be removed.
Having said that, great job!

PR #1519, opened 2020-11-05
[WDFT-AMST102020] - MAYA AUSSET

MayaAusset:


views/beers.hbs
```diff
@@ -0,0 +1,16 @@
+
+<body>
```
LloydC, line 2: You don't need a <body> here as every template engine is going to be rendered as part of the 'layout.hbs' inside {{{body}}}

app.js
```diff
@@ -22,4 +22,24 @@ app.get('/', (req, res) => {
...
+  punkAPI
+  .getRandom()
+  .then((responseFromAPI) => {
+     res.render('random-beer.hbs', { responseFromAPI });
+    })
+  .catch(error => console.log(error))
+});
+
+
+app.listen(3000, () => console.log('🏃running on port 3000'));
```
LloydC, line 44: It's considered good practice to have a callback to console.log which port your server is running on, I also recommend you get into the habit of storing your port number in a variable (it's commonly done as well)
```
const PORT = 3000;

app.listen(PORT, () => console.log(`Server is running on port ${PORT}`));

PR #1525, opened 2020-11-08
[Remote WD PT 081120] Thomas Schweers

t4ms:
The partials do not work - don´t know why?!

### DIFF
--- a/views/beers.hbs
+++ b/views/beers.hbs
@@ -1,0 +1,17 @@
+<div class="container">
+  <h1>Beers</h1>
+  <div class="row">
+        {{#each beersFromApi}} 
+        <div class="col-md-4">
+        <div class="card">
+            <img class="card-img-top" id="beerimg" src="{{this.image_url}}" alt="{{this.name}} {{this.lastName}}"/>
+            <div class="card-body">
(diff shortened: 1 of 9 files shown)


views/partials/beersArray.hbs
```diff
@@ -0,0 +1,14 @@
+{{!-- Did not work --}}
```
In-S4n3, line 1: Hey brother, this looks ok, so the problemas isn't here!

views/beers.hbs
```diff
@@ -0,0 +1,17 @@
+<div class="container">
+  <h1>Beers</h1>
+  <div class="row">
+        {{#each beersFromApi}} 
+        <div class="col-md-4">
```
In-S4n3, line 5: Try to comment this code which you already have in the partial and use this:

{{> beersArray this}}

views/random-beers.hbs
```diff
@@ -0,0 +1,15 @@
+<div class="topnav">
+  <h1>Random Beer</h1>
+  {{#each randomBeer}}
+  <div class="col-md-4">
```
In-S4n3, line 4: {{> beersArray this}} use this

PR #1528, opened 2020-11-09
[Remote-PTWD-092020] Daina Slezaite

daina-slezaite:
Not sure why, but css styling doesn't work for pages that use partials...

### DIFF
--- a/views/layout.hbs
+++ b/views/layout.hbs
@@ -6,7 +6,7 @@
     <link rel="stylesheet" href="/stylesheets/styles.css" />
   </head>
   <body>
-    <nav>
+    <nav class="navbar">
         <a href="/">Home</a>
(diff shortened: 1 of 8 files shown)


views/layout.hbs
```diff
@@ -0,0 +1,16 @@
...
+    <title>IronBeers</title>
+    <link rel="stylesheet" href="/stylesheets/styles.css" />
+  </head>
+  <body>
+    <nav class="navbar">
+        <a href="/">Home</a>
+        <a href="/beers">Beers</a>
+        <a href="/random-beer">Random Beer</a>
+    </nav>
+    {{{ body }}}
```
In-S4n3, line 14: You could add a footer here! 👍

Besides that, everything looks great. You can just make it look nicer and delete the commented code.

PR #1570, opened 2020-11-11
[WDFTRM-2020-11-11] - Gosia 

GosiaMas:
ugly but finished lab
I struggled a lot with implementation of API in route/path


IH-WebDev-TA-Remote:
Well done! All basic iterations done and good use of partials. Code is clean and easy to read. Check more about the usage of params used to solve the last bonus iteration as working with params will be very important in your projects. Keep up the great work! :)

PR #1573, opened 2020-11-11
[WDFTRM-20201111] - Pauline Massé

paulinemss:
All done, it was fun :)


IH-WebDev-TA-Remote:
Very nice job! all iterations done and everything working correctly. Also very neat simple styles with lots of flexbox :) Keep up the great work!

PR #1588, opened 2020-11-12
BERLIN, Vesna

ararita:


app.js
```diff
@@ -44,6 +44,7 @@ app.get('/beers', (req, res) => {
     punkAPI
       .getRandom()
       .then(random => {
+        // console.log('---------random', random);
         res.render('random-beer', { random });
       })
       .catch(error => console.log(error));
```
pkevinklein, line 50: Great job Vesna! Keep it up! When you have some time you can go over the bonuses. I'm sure they will be no brainer for you ;)

PR #1589, opened 2020-11-12
Berlin-WDFT-Oct2020-Davide-Andreani

dialmojordan:
Main iterations done


timbotimber (commented):
This looks great! Were you happy with the progress you made? Do you understand how the partials work?

dialmojordan:
They are templates separated in their own folder that I can recall  in
other templates by using handlebars syntax. However, I need to practice a
lot more because I’m still wasting too much time on them

*Handlebars* allows for template reuse through *partials*. *Partials* are
normal *Handlebars* templates that may be called directly by other
templates.Feb 25, 2020

PR #1591, opened 2020-11-12
Berlin - FToct20 - Iriel Aloni - Beer lab

irieljm:
partly done. too much errors

### DIFF
--- a/app.js
+++ b/app.js
@@ -9,7 +9,7 @@ const punkAPI = new PunkAPIWrapper();
 
 app.set('view engine', 'hbs');
 app.set('views', path.join(__dirname, 'views'));
-
+hbs.registerPartials(path.join(__dirname, 'views/partials'));
 app.use(express.static(path.join(__dirname, 'public')));
 
 // Register the location for handlebars partials here:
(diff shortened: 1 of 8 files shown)


views/beers.hbs
```diff
@@ -0,0 +1,5 @@
+ <div class="beers-container">
+        {{#each beerList}}
```
Jeff-Moraes, line 2: here instead of using the `beerList` you should use `beers` as it is the name of the propriety that you are sending to this view.

app.js
```diff
@@ -22,4 +22,24 @@ app.get('/', (req, res) => {
...
+    console.log(beerList)
+    res.render('beers', {beers: beerList})
+  }).catch(error => {
+    console.log("error in beers: ", error)
+  })
+  })
+
+app.get('/random-beers', (req, res) => {
+  punkAPI.getRandom().then((responseFromAPI) => {
+    res.render('random-beers', {randomBeer: responseFromApi});
```
Jeff-Moraes, line 36: here you are not sending the correct data that you are getting from the API.
`responseFromAPI` !== `responseFromApi`
remembering that javascript is case sensitive

another thing is the name of the view `random-beers` that is different from the view that you have `random-beer`

views/random-beer.hbs
```diff
@@ -0,0 +1,18 @@
+
+<div>
+    <div class="beers-container">
+        {{#each radomBeer}}
+            {{> beerpartial this}}
```
Jeff-Moraes, line 5: the name of the partial here `beerpartial` is different from the name of the partial file `beerPartial`

Jeff-Moraes (commented):
Well done Iriel! The logic and structure are all correct. :)
Most of the errors that you had are related to mistyped words.

PR #1594, opened 2020-11-13
mandatory done

WilliamCampbellie:


app.js
```diff
@@ -22,4 +22,24 @@ app.get('/', (req, res) => {
...
+  punkAPI.getBeers()
+  .then(response => res.render ('beers', {response}) )
+});
+// console.log(response);
+
+app.get('/random-beers', (req, res) => {
+  punkAPI.getRandom()
+  .then(response => res.render ('random-beers', {response}) )
+});
+
```
pkevinklein, line 35: Great job William! The only thing I noticed is the naming for the returning value of the promises (repsonse) is repetitive. You would like to give it a more descriptive name like "randomBeer" or "beersList" to make it more readable. Whenever you have some time, I suggest you to go through the bonuses :)

views/random-beers.hbs
```diff
@@ -0,0 +1,12 @@
+<h1>Have random Beer!</h1>
+
+    <div >
+      {{#each response}}
+  <img src="{{image_url}}" alt="Poster for {{title}}">
+        <p>{{this.name}}</p>
+        <p>{{description}}</p>
+        <p>{{this.tagline}}</p>
+        <p>{{this.food_pairing}}</p>
```
pkevinklein, line 9: Here I would recommend you to loop over the food_pairing array ( {{#each}} ) so that you have more control over displaying the data within that array. Keep up the good work!

PR #1597, opened 2020-11-13
BER-OCT2020- WEBDEV-Agus Garro

agusgarro:
well it happened to me that I can log the localhost:3000/ but not the /beers, /random-beer. I don't know if its a typo or what i did wrong. At the beginning I managed to make it appear but then i have no idea what I changed for it not to work. 
It happened the same in the first lab the express one, and I fail to find where the error is so I can fix it.

### DIFF
--- a/app.js
+++ b/app.js
@@ -13,44 +13,34 @@ app.set('views', path.join(__dirname, 'views'));
 app.use(express.static(path.join(__dirname, 'public')));
 
 // Register the location for handlebars partials here:
-hbs.registerPartial(path.join(__dirname, 'views/partials'));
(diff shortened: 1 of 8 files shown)


timbotimber (commented):
This all looks great. The logic is sound and it's great you used the partials and styled it. Did you get the routes to work? We can review it if you like when you have some time.

agusgarro:
I could make the first route work '/' as you can see. The other ones I got them to work but I DONT KNOW WHAT I DID and then they didn't work anymore.

agusgarro:
the css of the second and third route was never working though.

PR #1600, opened 2020-11-14
Berlin WDFT Oct 2020 Anna Waszak

annawaszak:
I don't know why the clicked beer page doesn't look the same as random beer, they should be using the same handlebars template.


timbotimber (commented):
Cool you used Async Await! Looks so much neater doesn't it. Nice you attempted the stylings too.

PR #1601, opened 2020-11-15
[WDFT BER OCT2020] Jan Völker 

jan-webdev:
@TAs: I had the strange problem (maybe bug?) that my partial 'beerCard' is being recognized and used but I can't change it anymore. Whenever I changed something in the partial's file, it didn't get put into the HTML site. As if hbs would be caching an older version of beerCard.hbs.
I tried:
- Changing things (like the order of items to be displayed
- Restarting node
- Refreshing the site
- Deleting the partial and creating it again

Nothing had any effect. Do you have any idea what's happening?
I wanted to do the iteration with re-using the partial, but didn't work out as I couldn't change it the way I needed it to change.

### DIFF
--- a/app.js
+++ b/app.js
@@ -21,7 +21,13 @@ app.get('/', (req, res) => {
   res.render('index');
 });
 app.get('/beers', (req, res) => {
-  res.render('beers');
+  punkAPI
+  .getBeers()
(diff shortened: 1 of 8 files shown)


Jeff-Moraes (commented):
I changed the partial file and restarted the node, and I can see the change in the browser, so for me here it is working.
I don't know why this was happening with your version, as it is working for me.
The lab is very well done! Great job Jan! :)

PR #1602, opened 2020-11-15
WDFT OCT 2020 BER Hannah Senft

Git-Hannah:
as far as possible


timbotimber (commented):
Hey looks like you were on the right path here but you were struggling a bit with passing the information to the front end? Do you feel better about this after the following lectures? We can review if you like.

PR #1606, opened 2020-11-18
WebDev Berlin FelixF 

FunkeFelix:


timbotimber (commented):
Nice one man, looking good. Did you console.log before you pass the info to the front end in the random beer route? Might help you work out what's going on there.

PR #1624, opened 2020-12-04
[PAR-WDFT-911] Kate Pavlova

katepavlova1:
done without styling and bonus

### DIFF
--- a/app.js
+++ b/app.js
@@ -14,12 +14,26 @@ app.use(express.static(path.join(__dirname, 'public')));
 
 // Register the location for handlebars partials here:
 
-// ...
+hbs.registerPartials(__dirname + "/views/partials");
 
 // Add the route handlers here:
 
 app.get('/', (req, res) => {
   res.render('index');
 });
+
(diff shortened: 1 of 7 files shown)


VProum:
Hi Kate, 
well done but no CSS and i think it is because you didn't manage to link your css file to the html.
same issue with the static image.

app.use(express.static(path.join(__dirname, 'public'))); on the js file tells your computer to use the public folder as start

so your path should not contain "../public/images/beer.jpg" but "/images/beer.png" instead

When you to a href on your links, it is better to start with a "/random-beer" instead of "random-beer"

| properties/ page | `/`      |        `/beers`      |   `/random-beer`   |
| :--------------: | :----------------: | :----------------: | :----------------: |
|      routes      | :white_check_mark: | :white_check_mark: |  :white_check_mark: |
|       HBS    | :white_check_mark: | :white_check_mark: | :white_check_mark: |
|   CSS    | :x: | :x:  | :x:  |

PR #1628, opened 2020-12-04
[PAR-WDFT-911] TRAN Thi Thanh Xuan

tranttxuan:


VProum:
Hi Xuan,

Nice job overall, you even do the CSS in responsive !

you have one thing that is not optimised at all, it is the route for beers.

```
 //fetch data of 25 beers
  const listBeers = [];

  const numb = 26;

  for (let i = 1; i < numb; i++) {
    await punkAPI
      .getBeer(i)
      .then(beersFromApi => listBeers.push(beersFromApi[0]))
      .catch(error => console.log(error, i));

  }
```
Why do you create a special const numb just for the for loop? 
 for (let i = 1; i < 26; i++) will do the job

but furthermore, why do you need a for loop?
it is unnecessary and will take some time because you are doing 25 calls to the API.
if you just use getbeers() :

```
app.get('/beers', async (req, res) => {
  const allBeers = await punkAPI.getBeers();
  res.render('beers', {allBeers});
})
```

your website will be faster as it only call the API once.

Last detail, you have to choose between async - await OR do a .then().catch()
don't do both it is overkill :D


| properties/ page | `/`      |        `/beers`      |   `/random-beer`   |
| :--------------: | :----------------: | :----------------: | :----------------: |
|      routes      | :white_check_mark: | :x: |  :white_check_mark: |
|       HBS    | :white_check_mark: | :white_check_mark: | :white_check_mark: |
|   CSS    | :white_check_mark: | :white_check_mark:   | :white_check_mark:   |

PR #1630, opened 2020-12-04
[PAR WEBDEV 911] Calypso Asline

Calyaln:


Fanny-Sbn:
Hello Calypso, thanks for submitting your lab. 

Nice job overall :D but there is no CSS :( 
You add some info in random-beers.hbs after the partial, you could have done an if else statement directly in the partial checking if you're on the /beers page or the /randombeer page and display the info accordingly (but it's not mandatory)

Wax on, wax off!

PR #1632, opened 2020-12-04
[PAR-WDFT-911] - Thomas LAMY

thomasl4012:
Lab done with bonuses

### DIFF
--- a/app.js
+++ b/app.js
@@ -11,6 +11,7 @@ app.set('view engine', 'hbs');
 app.set('views', path.join(__dirname, 'views'));
 
 app.use(express.static(path.join(__dirname, 'public')));
+hbs.registerPartials(path.join(__dirname, 'views/partials'));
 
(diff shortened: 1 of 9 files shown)


Fanny-Sbn:
Hello Thomas, thanks for submitting your lab. 

It's super lit, you even done the CSS, so smooth to look at it. 

Two comments : 

-Checking for your random beer it's not optimized. Since it's only one beer, you could send as a view parameter the only beer that you need and then call it in your randomBeer view.

app.get('/random-beer', async (req, res) => {
  try {
    const randomBeers = await punkAPI.getRandom();
    res.render('random-beer', { randomBeers: randomBeers[0] });
  } catch (err) {
    console.log(err);
  }
});

-Just try not to inline css style inside your hbs files or at one point you won't remember where you put your styles or which style is applying (css files or inline hbs?)

Keep on doing an amazing job 💯

PR #1633, opened 2020-12-04
[PAR-WDFT-911]Mina Samwail

MinaSamwail:
lab done 0 to 4.1

### DIFF
--- a/app.js
+++ b/app.js
@@ -22,4 +22,25 @@ app.get('/', (req, res) => {
   res.render('index');
 });
 
+app.get('/beers', (req, res) => {
+  punkAPI
+    .getBeers()
+    .then(beersFromApi => {
+      console.log('Beers from the database: ', beersFromApi);
+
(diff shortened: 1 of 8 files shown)


Fanny-Sbn:
Hello Mina, thanks for submitting your lab. 

It's super good job just a few comments:

-Checking for your random beer it's not optimized. Since it's only one beer, you could send as a view parameter the only beer that you need and then call it in your randomBeer view.

app.get('/random-beers', (req, res) => {
  punkAPI
    .getRandom()
    .then(randomBeer => {
      res.render('random-beers', { randomBeer: randomBeer[0] });
    })
});

You make an effort for the CSS, thaaaank youuuu hahaha

Last comment : I'm sure you're able to do bonus 6, give it a go ;) 

Keep on doing an amazing job Mina 💯

MinaSamwail:
Thanks for the feedback Fanny it will be useful !!! see ya

PR #1636, opened 2020-12-06
[PAR-WDFT-911] Lucas Souleyrasse

LucasSlrs:
all done but css is not finished, cannot find how to put 3 beers next to each others

### DIFF
--- a/app.js
+++ b/app.js
@@ -14,12 +14,30 @@ app.use(express.static(path.join(__dirname, 'public')));
 
 // Register the location for handlebars partials here:
 
-// ...
+hbs.registerPartials(path.join(__dirname, 'views/partials'));
 
 // Add the route handlers here:
 
 app.get('/', (req, res) => {
   res.render('index');
 });
+app.get('/beers', async (req, res) =>{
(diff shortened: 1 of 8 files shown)


Fanny-Sbn:
Hello Lucas, thanks for submitting your lab. 

Nice job for the mandatory iterations and nice effort for the CSS. 

Here are a few comments :

Checking for your random beer it's not optimized. Since it's only one beer, you could send as a view parameter the only beer that you need and then call it in your randomBeer view.
app.get('/random-beers', async(req, res) => {
  await punkAPI
  .getRandom()
  .then(responseFromAPI => {
    res.render('random-beers', { responseFromAPI: responseFromAPI[0]});
  })
});

For the CSS part, to have your beers next to each other you need to use flexbox, you have to wrap your "each" with a div and apply a display: flex and use the proper properties (flex-direction, justify-content, flex-wrap, etc...) 

Finally your link to display the information of a specific beer doesn't work, you haven't done the route and create a proper view, I'm sure you can do it even if it's a bonus. 

Wax on, wax off! :)

PR #1637, opened 2020-12-06
[PAR-WDFT-911]  Milande Teisserenc-bonestève

milly-dev:


Fanny-Sbn:
Hello Milande, thanks for submitting your lab. 

Nice for the mandatory iterations, but there are no bonuses :( 

Checking for your random beer it's not optimized. Since it's only one beer, you could send as a view parameter the only beer that you need and then call it in your randomBeer view.

app.get('/random-beer', async (req, res) => {
try {
const randomBeers = await punkAPI.getRandom();
res.render('random-beer', { randomBeers: randomBeers[0] });
} catch (err) {
console.log(err);
}
});

Wax on, wax off!

PR #1783, opened 2021-04-13
[FTWDAMS032021] - Jamal Nasser

jamal-nasser:


views/beers.hbs
```diff
@@ -0,0 +1,10 @@
+<h1>Beers</h1>
+<div class="beer-container">
+
+{{#each beersFromApi}}
+<img src="{{this.image_url}}" alt="beer">
```
reMRKableDev, line 5: Some beers do not have images, there could be some logic here that checks for this

reMRKableDev (commented):
Overall, great stuff @jamal-nasser 🚀 

I left a comment on a piece of logic you can think of to add

PR #1811, opened 2021-04-15
BER WDFT MAR2021 - Olesia Kulykova (didn't finish, no css)

olakul:


wax-bear (commented):
Great stuff Ola!
The code looks good, works as expected and as you indicated in the commit message, there is no CSS yet.
You could think about rendering the beers in a grid, and styling the page in general according to the example.

PR #1812, opened 2021-04-15
[BER-FT-032021] Philipp Dam

Philipphub:


app.js
```diff
@@ -14,12 +15,29 @@ app.use(express.static(path.join(__dirname, 'public')));
...
+    })
+    .catch(error => console.log(error));
+});
+
+app.get('/random-beer', (req, res) => {
+  punkAPI
+    .getRandom()
+    .then(responseFromAPI => {
+      console.log(responseFromAPI);
+      res.render('random-beer', { beersDataRandom: responseFromAPI });
```
ahmedhany82, line 38: This is correct but you could also just pass `responseFromAPI[0] `to `res.render` since the `.getRandom() `api will always return an array of one element and so we can avoid the` #each` loop in the hbs.

views/random-beer.hbs
```diff
@@ -0,0 +1,9 @@
+<ul>
+    {{#each beersDataRandom}}
+    <li style="list-style-type: none;">
+        <img src="{{this.image_url}}">
+        <p>{{this.name}}</p>
+        <p>{{this.description}}</p>
```
ahmedhany82, line 6: You could give it a try adding food_pairings and brewers_tips.

views/beers.hbs
```diff
@@ -0,0 +1,11 @@
+<ul>
+    {{#each beersData}}
+    <li>
+        <h2>{{this.name}}</h2>
+        <h3>{{this.tagline}}</h3>
+        <img src="{{this.image_url}}">
+        <p>{{this.description}}</p>
```
ahmedhany82, line 7: Give it a try using a partial. You can write it once and reuse it here and in the random beer as well.

ahmedhany82 (commented):
Hi Philipp, good job! please find my review remarks for this lab. Feel free to reach out if you have any questions.
Best regards,
Ahmed

PR #1815, opened 2021-04-15
[FT -WD - BER] - David Bueno

davidbuenox8:


app.js
```diff
@@ -22,4 +22,25 @@ app.get('/', (req, res) => {
...
+    .catch(error => console.log(error));
+
+})
+
+app.get('/random-beer', (req, res) => {
+  punkAPI
+    .getRandom()
+    .then(randomBeer => {
+      console.log('Random beer from the database: ', randomBeer);
+      res.render('randombeer', { randomBeer })
```
ahmedhany82, line 41: You correctly handled this as an array in the hbs using #each. Since this will always be an array of 1 element you could alternatively pass ` randomBeer[0]` to res.render and then you only have a single element to handle in the hbs

views/beers.hbs
```diff
@@ -0,0 +1,7 @@
+<h1>Beers</h1>
+{{#each beersFromApi}}
+<h2>{{this.name}}</h2>
+<img src="{{this.image_url}}" alt="beer">
+<h3>{{this.tagline}}</h3>
+<p>{{this.description}}</p>
```
ahmedhany82, line 6: If you have some time, try experimenting with Partials. It is a great way to reuse code in different hbs files. See here we have `image`, `name`, `tagline` and `description` and they are used again in case of a random beer. These 4 could go into a Partial.

ahmedhany82 (commented):
Hey David,

Good job! Please find my review comments. Feel free to reach out in case you have any questions.

Regards,
Ahmed

PR #1821, opened 2021-04-15
first commit

cintiagarcia:


views/random-beer.hbs
```diff
@@ -0,0 +1,16 @@
...
+        {{#each randomBeer}}
+        
+        {{> beerpartial}}
+        
+        <ul>
+            {{#each food_pairing}}
+                <li>{{this}}</li>
+            {{/each}}
+        </ul>
+        <h2>{{brewer_tips}}</h2>
```
ahmedhany82, line 13: The brewer tips are not displayed. I think the name field is `brewers_tips`.

app.js
```diff
@@ -14,12 +14,33 @@ app.use(express.static(path.join(__dirname, 'public')));
...
+
+    .catch(error => console.log(error));
+});
+
+app.get('/random-beer', (req, res) => {
+  punkAPI
+    .getRandom()
+    .then(responseFromAPI => {
+      console.log(responseFromAPI.name);
+      res.render('random-beer', { randomBeer: responseFromAPI });
```
ahmedhany82, line 40: This is correct. Another way to do it is to pass` responseFromAPI[0] `to res.render. This is because the `.getRandom() `always return an array of one element and so we can skip having to do the loop in the hbs.

ahmedhany82 (commented):
Hi Cintia, good job for this lab. Please find my review comments.

Best regards,
Ahmed

PR #1822, opened 2021-04-15
[BER-FT March] Hélène & Andy

AndyMcGN:
CSS finished for random beer page

### DIFF
--- a/app.js
+++ b/app.js
@@ -43,4 +43,4 @@ app.get('/random-beer', (req, res) => {
 
 
 
-app.listen(3000, () => console.log('🏃‍ on port 3000'));
+app.listen(5000, () => console.log('🏃‍ on port 5000'));
--- a/app.js
+++ b/app.js
@@ -14,12 +14,33 @@ app.use(express.static(path.join(__dirname, 'public')));
 
 // Register the location for handlebars partials here:
 
-// ...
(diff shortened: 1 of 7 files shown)


app.js
```diff
@@ -14,12 +14,33 @@ app.use(express.static(path.join(__dirname, 'public')));
...
+    res.render('beers', {beersFromApi});
+  })
+});
+
+app.get('/random-beer', (req, res) => {
+  punkAPI
+  .getRandom()
+  .then(randomBeer => { 
+    console.log(randomBeer);
+    res.render('random-beer', {randomBeer}); //question: could we do randomBeer[0] to directly access the object?
```
wax-bear, line 40: Yes, I would just send the single and only element in the array to the front end.
In fact, I do not find it very logical that Punk API sends "random beer" as an array.

views/partials/beerpartial.hbs
```diff
@@ -0,0 +1,4 @@
+<img src="{{image_url}}" alt="beer image">
```
wax-bear, line 1: Sometimes the random beer does not contain an image URL, thus no image.
In the "/random-beer" route, the image_url property value can be set to a default image in case it is undefined.

wax-bear (commented):
Great stuff guys!
In terms of styling, you nailed it.
Please check out the comments for potential improvements.

PR #1823, opened 2021-04-15
BER - FTMA2021 - Jan Colomer

Jancololo:


views/random-beer.hbs
```diff
@@ -0,0 +1,15 @@
+<div class="random-beers-wrap">
+    <div class="random-beer-container">
+        {{> beerPartial randomBeer}}
+        {{#with randomBeer}}
```
annawaszak, line 4: You could include the beerPartial to be included within the "with", in such case you could have to change `{{> beerPartial randomBeer}}` to `{{> beerPartial this}}`

views/partials/beerPartial.hbs
```diff
@@ -0,0 +1,6 @@
+{{#if image_url}}
+<img class="beer-img" src="{{image_url}}" alt="{{name}} image">
+{{/if}}
```
annawaszak, line 3: Good job with the if statement here !

annawaszak (commented):
Hi Jan,

Outstanding job on the lab, I was hoping you would go for the last challenge and make each beer clickable to display the details. It would not be much work at this point, as you could use the "random-beer" view to display it :)

Thanks,
Anna

views/partials/beerPartial.hbs
Jancololo: Hi Anna! Thanks for the feedback! I've just finished the bonus iteration. I thought I had done it...I have committed the new changes but I'm not sure if you are able to see them in that pull request already or if I should create another pull request.

Cheers

PR #1824, opened 2021-04-15
[BER-WDFT-MAR21] Fintan Hahnefeld

fintanhahnefeld:


views/partials/beerpartial.hbs
```diff
@@ -0,0 +1,17 @@
+{{#each beersArray}}
```
annawaszak, line 1: this part `{{#each beersArray}}` should be a part of the beer.hbs file. Since it's supposed to display all the beers there, that's where you should iterate over them. The partial is only one of them.

views/beers.hbs
```diff
@@ -0,0 +1,2 @@
+
+    {{> beerpartial }}
```
annawaszak, line 2: This is where you should iterate over the beers. 
Wherever you use a partial you pass on the information you want to display there. Which means you will need to add something after the beerpartial for it to work.

views/partials/beerpartial.hbs
```diff
@@ -0,0 +1,17 @@
...
+
+<div class="beerscontainer">
+<div class="beers-content">
+    <div>
+        <img id="beerspage-img" src="{{image_url}}" alt="poster for {{name}}">
```
annawaszak, line 6: Some beers do not have images. In that case the user can see the little image icon and what you added in "alt". I would suggest to wrap this image tag in an if statement. That way, you can't see the image if there isn't any.

views/random-beer.hbs
```diff
@@ -0,0 +1,21 @@
...
+    <h6>Food Pairing:</h6>
+        {{#each randomBeer.food_pairing}}
+
+        <li>{{this}}</li>
+
+        {{/each}}
+    </ul>
+
+</div>
+</div>
```
annawaszak, line 21: You could wrap this entire code with a `{{#with randombeer}}`. That way you would not have to add randombeer everywhere you access it. e.g. instead of "{{ randomBeer.food_pairing}} "{{, just "food_pairing}}".

annawaszak (commented):
Hi Fintan,

Good job on the lab! You can find my notes as comments below.

Thanks,
Anns

PR #1825, opened 2021-04-15
[B-WDFT-MAR2021] Steffen Leppin

ThisOptimism:


wax-bear (commented):
Great stuff Steffen!
Overall the templates and styling look good.
One thing that could be added is bonus iteration 6:

"Make all the beers on the beers page clickable. If users click on a specific beer, they should be able to see a page with the detailed information of that particular beer"

This would also require using a dynamic route on the back end.

Cheers!

PR #1826, opened 2021-04-15
[Berlin-WDFT-MAR2021] Marcia Faria Netter

mfndnc:


public/stylesheets/styles.css
```diff
@@ -5,3 +5,119 @@ dark-blue: #0a2c42;
...
+  margin: 0;
+  padding: 0;
+  display: flex;
+  flex-direction: column;
+  height: 100vh;
+}
+header,
+footer {
+  flex: 0 0 content;
+  min-height: 4vh;
```
ahmedhany82, line 19: I think if you set the height here instead of min-height, it will be consistent when rendering different hbs files.

public/stylesheets/styles.css
```diff
@@ -5,3 +5,119 @@ dark-blue: #0a2c42;
...
+  font-family: Arial, Helvetica, sans-serif;
+  margin: 0;
+  padding: 0;
+  display: flex;
+  flex-direction: column;
+  height: 100vh;
+}
+header,
+footer {
+  flex: 0 0 content;
```
ahmedhany82, line 18: This causes a warning "Invalid property value" if you check it in the developer tools.

ahmedhany82 (commented):
Hey Marcia,

Great work! Just added one tiny comment for the height of the header. 

All the best!
Ahmed

PR #1827, opened 2021-04-15
[Berlin - Web Dev 2021] Sullivan biardeau

Marc-89:


wax-bear (commented):
Great stuff Sullivan!
I added some comments for potential improvements.
You could look to the style, further iterate on that.
Also, "random-beer" is not working in my machine. Is it in yours?

image

Cheers!

PR #1828, opened 2021-04-15
[BER WDFT 032021] Pietro Ballarin

pietroballarin:


app.js
```diff
@@ -22,4 +22,14 @@ app.get('/', (req, res) => {
...
 });
 
-app.listen(3000, () => console.log('🏃‍ on port 3000'));
+app.get('/beers', (req, res) => {
+  punkAPI.getBeers().then(beers => res.render('beers', {beerList: beers}))
+  .catch(error => console.log(error)); 
+});
+
+app.get('/random-beer', (req, res) => {
+  punkAPI.getRandom().then(beer => res.render('random-beer', {randomBeer: beer}))
```
wax-bear, line 31: 1. I would distribute this code over a few lines
2. You could send the single random beer, by retrieving it from the array, instead of sending an array to the front end.

public/stylesheets/styles.css
```diff
@@ -1,7 +1,68 @@
-/*
+/* /*
```
wax-bear, line 1: The CSS looks nice and clean.
Do you still need this comment?

wax-bear (commented):
Hey Pietro!

Question: did you create views, and if so, did you (not) commit them?
When running your project, I run into errors when going to "beers" and "random-beer".

After you commit the views, I can review the project in its entirety.

Best,

Thomas

wax-bear (commented):
Well done Pietro!

The functionalities work as expected and the code generally looks good.
You could think of styling the anchor tags in the nav bar and the buttons on the home page.
Also, the alignment of beers could be more central, or even in a grid.

Cheers,

Thomas

PR #1830, opened 2021-04-16
[BER-WDFT-MAR2021] Diogo Major

diogocmjr:


app.js
```diff
@@ -14,12 +14,30 @@ app.use(express.static(path.join(__dirname, 'public')));
...
+    res.render('beers', {beersFromApi});
+  })
+  .catch(error => console.log(error));
+});
+
+app.get('/randombeer', (req, res) => {
+  punkAPI
+  .getRandom()
+  .then(responseFromAPI => {
+    res.render('randombeer', {responseFromAPI});
```
ahmedhany82, line 38: This is correct. Another way to do it is to just pass `responseFromAPI[0]` to res.render since the `.getRandom()` will always return an array of one element and so we can skip having to use `#each` in the hbs.

ahmedhany82 (commented):
Hey Diogo, 
Great job! I only added one small review comment. Keep up the good work!
Best regards,
Ahmed

PR #1832, opened 2021-04-18
[WDFT-MAR21-BER]- Svjetlana Rados

SvjetlanaR:


views/random-beer.hbs
```diff
@@ -0,0 +1,6 @@
+{{> beerpartials this}}
+
+{{#each food_pairing}}
+{{this}}
+{{/each}}
+{{brewers_tips}}
```
ahmedhany82, line 6: You are passing randomBeer to this hbs and here randomBeer is an array of 1 element. So this is why it does not work. The api .getRandom() returns an array of 1 element and this you can check if you try to console.log the return from the API. 

One way to make it work is to wrap it with `{{#each randomBeer}}...{{/each}}` tags. A better way is to pass `randomBeer[0] `instead of `randomBeer` in line 39 in `app.js`. Then you do not need the` #each` in the hbs.

     ` res.render('random-beer', { randomBeer: randomBeer[0] } ))`

This is how the hbs can look:

 ```
   {{> beerpartials randomBeer}}

    {{#each food_pairing}}
        {{this}}
    {{/each}}
    {{brewers_tips}}
```

File: views/partials/beerpartials.hbs [hunk 2]
```diff
@@ -0,0 +1,6 @@
+<h1>Ironbeers</h1>
```
ahmedhany82, line 1: This Ironbeers header tag will appear everywhere and also for each beer. Better to move it out.

ahmedhany82 (commented):
Hi Svjetlana, 
Good job for the lab! Please find my review remarks. Feel free to reach out if you have any questions.
Best regards,
Ahmed

PR #1834, opened 2021-04-19
BerlinWD03-2021- Rimbert Richter

rimboriri:


views/random-beer.hbs
```diff
@@ -0,0 +1,20 @@
...
+        <ul>
+         {{#each this.food_pairing}}
+         <li>{{this}}</li>
+         {{/each}}
+         </ul>
+          <h3>Brewer Tips</h3>
+         {{brewers_tips}}
+
+       {{!-- <img src={{image_url}} alt="imgPic"> --}}
+    {{/each}}
```
annawaszak, line 18: An alternative solution to adding #each statement here, would be passing `beersFromApi[0]` to the view instead of `beersFromApi`. This is perfectly fine of course. 

But if you do it, you will have to replace `#each` with `#with`.

views/partials/beerpartial.hbs
```diff
@@ -0,0 +1,7 @@
+ <img src={{image_url}} alt="imgPic">
```
annawaszak, line 1: Since some beers do not have images and you can see the image icon with the alt text, you could consider wrapping this line in an if statement, so that we can only see the image if there is one.

views/random-beer.hbs
```diff
@@ -0,0 +1,20 @@
+<div>
+
+    {{#each randomBeer}}
+        {{!-- <h1>{{name}}</h1>
+      <h3>{{tagline}}</h3>  
+         {{description}} --}}
+          {{> beerpartial this}}
+           <h3>Food Pairing</h3>
+        <ul>
+         {{#each this.food_pairing}}
```
annawaszak, line 10: You can use "this" here if you'd like, but then I would be consistent and add it in front of "brewers_tips" too.

annawaszak (commented):
Hey Rimbo,

Great job! As always, I added a few comments with some improvement suggestions. If you have any questions, let me know. 

Thanks, 
Anna

PR #1835, opened 2021-04-19
done

Raffaeleb003:


views/partials/beerpartial.hbs
```diff
@@ -0,0 +1,19 @@
+<div class="beer-card">
+    <div class="image-container"><img src="{{this.image_url}}" alt="beer-photo"></div>
```
annawaszak, line 2: Since not all beers have images, I would recommend to wrap this line of code in an if statement so that the image only display it there is one. Currently it's displaying an image icon and the alternative text.

annawaszak (commented):
Hi Raffaele, 

Awesome job on the lab! You did all the bonuses! 

I left one little comment, but it wasn't anything mentioned in the instructions, just a tip.

Thanks, 
Anna

PR #1836, opened 2021-04-20
20 Apr 16:51

z-zeitler:


app.js
```diff
@@ -22,4 +27,25 @@ app.get('/', (req, res) => {
...
+     res.render('beers', {arrayObject: beersArray })
+     console.log(beersArray)
+    }).catch(error => {
+      console.log(error)
+    });
+});
+
+app.get('/random-beer', (req, res) => {
+  punkAPI.getRandom().then(randomBeer => {
+    res.render('random-beer', {beerObject: randomBeer})
```
ahmedhany82, line 42: randomBeer is an array of one object. So you could pass it to render like so randomBeer[0]. Also since it is an object by itself, you don't need to wrap it in an object.

res.render('random-beer', randomBeer[0])

ahmedhany82 (commented):
Hey Zoe, 
Please see my comment on the random beer below. I see you haven't tried the bonus part. Would encourage you to do so if you have some time.
All the best,
Ahmed

PR #1838, opened 2021-05-02
[PT-AMS-MAR21] Zsoka Montvai

ZsokaM:


app.js
```diff
@@ -17,9 +18,20 @@ app.use(express.static(path.join(__dirname, 'public')));
 // ...
 
 // Add the route handlers here:
-
 app.get('/', (req, res) => {
   res.render('index');
 });
 
+app.get('/beers', (req, res) => {
+  const beers = punkAPI.getBeers()
```
guido-an, line 26: You are not using the const beers here. Same on line 32.

You can simply start with `punkAPI.getBeers()`

ZsokaM: thank Guido! updated

PR #1907, opened 2021-06-24
BER-WDFT-MAY2021-Zhanna

Decdancer:


app.js
```diff
@@ -14,12 +14,33 @@ app.use(express.static(path.join(__dirname, 'public')));
...
+    .then(beersFromApi => res.render('beers', {AllB: beersFromApi}))
+    .catch(error => console.log(error));
+});
+
+
+
+app.get('/random-beer', (req, res) => {
+  punkAPI
+    .getRandom()
+    .then(responseFromApi => res.render('random-beer', {RandomB: responseFromApi}))
```
ahmedhany82, line 38: `responseFromApi` here is an array of one element. You could either pass the array as you did or you can also pass the beer object like so `{RandomB: responseFromApi[0]}`

views/random-beer.hbs
```diff
@@ -0,0 +1,28 @@
+<div><h1>Your Random Beer here! Cheers!</h1></div>
+
+{{#RandomB}}
```
ahmedhany82, line 3: This works but you can also use #each since you are passing an array (although an array of just one element). You can skip this and just use `RandomB` if you just pass the element instead of the array from app.js
Example:
`{{>beerpartial RandomB}}`
`{{#each RandomB.food_pairing}}`

If you are passing the element and want to set the scope once you could use `{{#with RandomB}}` and then you can do: 
`{{>beerpartial this}}`
`{{#each food_pairing}}`

ahmedhany82 (commented):
Hey Zhanna,
Good job working on this lab. I added a few remarks below. Perhaps if you have some time you could try the bonus part (iteration 6).
Regards,
Ahmed

PR #1909, opened 2021-06-24
[BER - WDFT - MAY2021] Hanlin Li

PemaHL:
iteration 7, css styling, not done yet

### DIFF
--- a/views/partials/beerpartial.hbs
+++ b/views/partials/beerpartial.hbs
@@ -1,0 +1,12 @@
+
+<div class = "beerPartial">
+    <div>
+    <a><img src={{this.image_url}} alt="beer" class = "image"></a>
+    </div>
+
+    <div>
+    <a href="/beers/{{this.id}}">{{this.name}}</a>
+    <a>{{this.description}}</a>
+    <a>{{this.tagline}}</a>
+    </div>
+</div>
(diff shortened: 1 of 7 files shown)


views/partials/beerpartial.hbs
```diff
@@ -0,0 +1,12 @@
+
+<div class = "beerPartial">
+    <div>
+    <a><img src={{this.image_url}} alt="beer" class = "image"></a>
+    </div>
```
atkat, line 5: You are wrapping the image in a link but there is no href, so the images are not clickable yet.

views/partials/beerpartial.hbs
```diff
@@ -0,0 +1,12 @@
...
+
+    <div>
+    <a href="/beers/{{this.id}}">{{this.name}}</a>
+    <a>{{this.description}}</a>
+    <a>{{this.tagline}}</a>
+    </div>
+</div>
```
atkat, line 12: A small thing, but it would be nice to have the info introduced, as opposed to just pasted in there. So for instance:

<p> Brewer's tips: {{ brewers_tips}} </p>

atkat (commented):
Hey Hanlin,

very good! There's not much for me to say, I've added just a couple of comments below. 

Let me know if you have any questions.

Best,
Katka

PR #1912, opened 2021-06-24
[BER-WDFT-MAY2021] Alexandre Oliveira

alexandreoliv:


views/partials/beerPartial.hbs
```diff
@@ -0,0 +1,5 @@
+<img id="beer-image" src="{{this.image_url}}" alt="Beer {{this.name}}">
+                        <div id="beer-info">
+                            <div><a href ="/{{this.id}}"><span id="beer-name">{{this.name}}</span></a></div>
+                            <div>{{this.description}}</div>
+                            <div>{{this.tagline}}</div>
```
atkat, line 5: You can close your div here, why did you close in the beers.hbs, was the styling messing up?

atkat (commented):
Hey Alexandre, well done, I have nothing to comment on!! :)

views/partials/beerPartial.hbs
alexandreoliv: Yes :(. random-beer.hbs and unique-beer.hbs have different if statements and styles to that of beers.hbs. I'm sure there are better ways to use partials in this lab but I focused mostly on the logic and the presentation. I added partials to my own backlog though :D. Would you recommend anything?

atkat: Oh, I changed my mind and I was gonna delete this comment but then I forgot. You're actually not wrong doing it this way. There's probably a better way to do it but I ended up doing the same in my lab :). Hm, sorry, I don't have any special recommendations I am afraid.

PR #1915, opened 2021-06-25
[WDFT - BER MAY2021] Raoul Ullens

Ra-Ul-1:
add iterations 1-4

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,3 +1,4 @@
+const { response } = require('express');
 const express = require('express');
 
 const hbs = require('hbs');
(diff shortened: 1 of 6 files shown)


views/random-beer.hbs
```diff
@@ -0,0 +1,5 @@
+<h1>Random Beer</h1>
+{{theBeer}}
```
ahmedhany82, line 2: I think this is not needed here. If you want to set the context, you could use `{{#with theBeer}}` and close it with `{{/with}}` but here you don't need to.

app.js
```diff
@@ -1,3 +1,4 @@
+const { response } = require('express');
```
ahmedhany82, line 1: I think this line is not needed. Could be an autocomplete by VS Code.

views/home.hbs
```diff
@@ -0,0 +1,2 @@
+{{!-- render index.hbs --}}
+{{> index.hbs}}
```
ahmedhany82, line 2: I think `home.hbs` is not rendered by any route and could be removed ?

ahmedhany82 (commented):
Hey Raoul,
Good work so far! Try to make sure you are comfortable with partials and perhaps give the bonus iteration (clickable beers) a try. Reach out if you have any questions.
Regards,
Ahmed

PR #1920, opened 2021-06-28
done

elihinojos:


app.js
```diff
@@ -22,4 +28,35 @@ app.get('/', (req, res) => {
...
+  .then(responseFromApi => {
+
+    res.render('random-beer', {randomBeer: responseFromApi});
+  })
+  .catch(error => console.log(error));
+});
+
+app.get('/beer/:id', (req, res) => {
+res.render('beer-details')
+}
```
atkat, line 57: you were missing a bracket in line 57 - it should be: })

views/partials/beerpartial.hbs
```diff
@@ -0,0 +1,10 @@
+<img src='{{image_url}}' alt='{{name}}'>
+
+<a href="/beers/beer-i3f4d34s34b"><img src="{{image_url}}" alt=""></a>
+
```
atkat, line 4: you have the image twice here :)

views/random-beer.hbs
```diff
@@ -0,0 +1,13 @@
...
+  <ul>
+    {{#each this.food_pairing}}
+        <li>{{this}}</li>
+    {{/each}}
+  </ul>
+  <p>
+    {{brewers_tips}}
+  </p>
+{{/each}}
+
```
atkat, line 13: A small thing, but it would be nice to introduce what that text is: <p> Food pairing: </p> and then <ul> etc.

views/partials/beerpartial.hbs
atkat: Also, I will assume you have just started with a hard coded url id and then didn't have enough time to finish. Let me know if you have any questions about this.

atkat (commented):
Hey Elizabeth,

Please see some comments below. 

Styling was not a priority in this lab, but it would be nice to have the nav bar a bit neater. 

Otherwise well done! :)

K

PR #1922, opened 2021-06-28
BER-WDFT-MAY21- Alan Sanchez

alnwolff:
Done with iterations.
Some styling is missing

### DIFF
--- a/views/partials/beerpartial.hbs
+++ b/views/partials/beerpartial.hbs
@@ -1,0 +1,5 @@
+<img src='{{image_url}}' alt='{{name}}'>
+
+{{name}}
+{{description}}
+{{tagline}}
--- a/views/random-beer.hbs
+++ b/views/random-beer.hbs
@@ -1,0 +1,8 @@
+<h1>RANDOM BEER</h1>
+
+{{#each randomBeer}}
+    {{> beerpartial this}}
+    {{food_pairing}}
+    {{brewers_tips}}
+
+{{/each}}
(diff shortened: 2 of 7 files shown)


views/partials/beerpartial.hbs
```diff
@@ -0,0 +1,5 @@
+<img src='{{image_url}}' alt='{{name}}'>
+
+{{name}}
+{{description}}
+{{tagline}}
```
atkat, line 5: You forgot to wrap the content in the respective tags: <p>{{name}}<p> etc.  This way only some of the text appears but all of it plus it's not actually added to the DOM (try to inspect it).

views/random-beer.hbs
```diff
@@ -0,0 +1,8 @@
+<h1>RANDOM BEER</h1>
+
+{{#each randomBeer}}
+    {{> beerpartial this}}
+    {{food_pairing}}
+    {{brewers_tips}}
```
atkat, line 6: same thing here

views/random-beer.hbs
```diff
@@ -0,0 +1,8 @@
+<h1>RANDOM BEER</h1>
+
```
atkat, line 2: Here, I think it would be nicer to have the title of the beer as the title of the page, as opposed to 'random beer'. We already know we chose a random one.

atkat (commented):
Hey Alan,

Nicely done, as always. Just make sure you add those html tags inside your hbs :)

PR #1923, opened 2021-06-30
BER-WDFT-Hiler Mateo

hiler-hutter:
Hi,
I'm trying still to do the bonus 6 and 7 and also some css styling. I'm not fast but I'm trying to understand well.

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,7 +1,9 @@
 const express = require('express');
 
 const hbs = require('hbs');
+
 const path = require('path');
+
 const PunkAPIWrapper = require('punkapi-javascript-wrapper');
 
 const app = express();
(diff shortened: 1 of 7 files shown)


views/partials/beerpartial.hbs
```diff
@@ -0,0 +1,18 @@
...
+    {{name}}
+</h2>
+
+<p>
+    {{description}}
+</p>
+
+<p>
+    {{tagline}}
+</p>
```
atkat, line 18: A small thing, but it would be nice to have the info introduced, as opposed to just pasted in there. So for instance:

Brewer's tips: {{ brewers_tips}}

app.js
```diff
@@ -14,12 +16,35 @@ app.use(express.static(path.join(__dirname, 'public')));
...
+app.get('/random-beer', (req, res) => {
+  punkAPI
+  .getRandom()
+  .then(responseFromAPI => {
+    console.log('Random beer ', responseFromAPI);
+    res.render('random-beer', {responseFromAPI});
+  })
+  .catch(error => console.log(error));
+})
+
```
atkat, line 49: You can remove the console logs once you're ready to submit :)

atkat (commented):
Hi Hiler, 

All good, I just have a couple of small comments, see below. And let me know if you have any questions.

Best,
Katka

PR #1933, opened 2021-07-26
[WDFT 286] Jessie Panganiban

jessievp:
routes and pages created, missing data from api


jessievp:
will work on css the next coming days!

PR #1955, opened 2021-08-31
[MAD DE] Vanessa Zagone

SybilVane:
All iterations done, apart from 7. 

I don't get how to use partials if I want to produce a col-4 in the beers page, and a col-12 in the random beer page. 

Is it possible to complete number 7 without using partials?

### DIFF
--- a/app.js
+++ b/app.js
@@ -14,7 +14,7 @@ app.use(express.static(path.join(__dirname, 'public')));
 
 // Register the location for handlebars partials here:
 
-// ...
+hbs.registerPartials(path.join(__dirname, 'views/partials'));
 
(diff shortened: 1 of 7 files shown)


ta-web-mad:
Hi there Vane,

By '7' I guess you mean the 6th iteration of the exercise, as it's the one you have left undone. To answer your question, allow me to ask you a new one: have you thought of letting the 'col' classes out of the partial? 
This would allow you to use the same partial in the beers page,  and in both the random beer page and the 'beer details' page (iteration 6).

To end, I have to congratulate you on your overall exercise resolution. Very well done!

PR #2066, opened 2021-11-11
[FT-LIS-102021] Karina-Giorgi

karinaglf:


app.js
```diff
@@ -13,13 +13,55 @@ app.set('views', path.join(__dirname, 'views'));
...
+
+});
+
+app.get('/random-beers', (req, res) => {
+  punkAPI
+  .getRandom()
+
+  .then(beersFromApi => {
+    console.log(beersFromApi)
+    res.render('random-beers',{beersFromApi});
```
stefanopisani, line 58: You can also extract the object from the array to have a cleaner code but it works perfectly also like this ;)

stefanopisani:
good job, keep going guys!!

PR #2118, opened 2022-01-15
JoaoNacho_AnaModesto_PTWebDev_Lisbon_Nov2021

joaonacho:
on iteration 3

### DIFF
--- a/app.js
+++ b/app.js
@@ -22,4 +22,19 @@ app.get('/', (req, res) => {
   res.render('index');
 });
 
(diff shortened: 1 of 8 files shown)


carlarsmendes:
Good job guys! Dont forget to: Always indent your html / code in general for readibility;
`hbs.registerPartials(path.join(__dirname, 'views/partials'));`--> this should be more on top of the page, not down below. Get used to organize your files  since they will grow quickly!

PR #2189, opened 2022-02-02
WD Remote Jan. 2022 - Ana & Aleksandra

pavaoana:


TA-Remote:
Great effort on the lab! I will be uploading solutions, but we can solve it together after class if that will help you understand better! K

PR #2191, opened 2022-02-02
basic tasks done

throwoutofcoffeeexception:


views/beers.hbs
```diff
@@ -0,0 +1,8 @@
+<div>
+    {{#each beer}}
+        {{this.name}}
+        <img src="{{this.image}}">
+        {{this.description}}
+        {{this.tagline}}
+    {{/each}}
+</div>
```
TA-Remote, line 8: - Here inside `{{#each}}` you are printing information of one beer, with every iteration. You want to wrap that information of each beer with `<div>`.

At the moment you have one div, and all the beers alltogether inside that div

- You are not printing images, cause you have wron property name. Always console.log data you receive to see exact name : ``

PR #2268, opened 2022-05-04
[WDFT0422] Ana Pinho

AnaGonPinho:


webLisbonTA:
Good job on completing the lab, on the getBeer() you should also render a description for each beer. Good job!

PR #2271, opened 2022-05-04
[FTLIS042022] Mariana Oliveira

miloliveira:
Lab done, plus bonus 5. 
Css was really hard specially in random beers part, because I was styling the partial file and that affects both beers and random beers.


webLisbonTA:
Well done Mariana!
yes I totally get you, it is tricky to style because the styles are different on both files, in that case you could work with the specificity of css to select only the random beer view

PR #2275, opened 2022-05-04
[FTLISAPR22] PEDRO COSTA

pedrocncosta:
DONE

### DIFF
--- a/views/randombeer.hbs
+++ b/views/randombeer.hbs
@@ -1,8 +1,10 @@
-<div >
+<div class='randomDiv'>
   {{#each resApi}}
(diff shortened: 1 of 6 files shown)


views/randombeer.hbs
```diff
@@ -0,0 +1,10 @@
+<div class='randomDiv'>
```
webLisbonTA, line 1: To complete the 4.2 random-beer iteration you should also add a food pairing list on the beer details.
A hint to implement that: if you take a look on the api response food_pairing is an array of string so you can look through that on the view to show each of the items on the array.

PR #2282, opened 2022-05-04
[FTLIS042022] matildeCosta

matildedacosta:
At first I used the bootstrap example just to do the iterations and when I went to change it to the one with css it disconfigured in all the other pages :/ I tried to fix it but I wasn't able to. I thought it may have something to do with the fact that I am adding the css files outside of the layout? But I didn't know how to do it while using the layout (and it was working with bootstrap)

### DIFF
--- a/app.js
+++ b/app.js
@@ -12,6 +12,8 @@ app.set('views', path.join(__dirname, 'views'));
 
 app.use(express.static(path.join(__dirname, 'public')));
 
+hbs.registerPartials(path.join(__dirname, 'views/partials'));
+
 // Register the location for handlebars partials here:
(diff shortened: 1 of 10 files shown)


webLisbonTA:
Hi Matilde, 
I got it! Your stylesheets are perfectly setup, the issue here is that you have used the same classes on different views and they are overriding each other.

For example on your layout you have classes for each one of your nav links, like:
```
        <a class='beers' href='/beers'>Beers</a>
```

and later on you have a beers view which also have a class `beers` on some elements. So the stylesheet for the beers view is overriding the styles of the navbar.

PR #2285, opened 2022-05-05
[FTAPR2022] João Morgado

JoaoMorgado8:


webLisbonTA:
Well done João!

To complete the 4.2 random-beer iteration you should also add a food pairing list on the beer details.
A hint to implement that: if you take a look on the api response food_pairing is an array of string so you can look through that on the view to show each of the items on the array.

PR #2295, opened 2022-05-31
lab-ironbeers

zarahsara:


Enes-Ironhack:
You have the correct views, but you do not have any js file :'( ...

PR #2297, opened 2022-05-31
Finished iteration 1 and 2;

nataliabuzila:
WIP iteration 3;
not started iteration 4


Enes-Ironhack:
Honestly, quite impressive. Good understanding of the API and the rendering. You are more than able to do itération 4 when you will have time. Good job

PR #2298, opened 2022-05-31
with problems

EvgeniiaRiaguzova:


Enes-Ironhack:
You actually use the API and the Layout correctly,  but for some reason you do "  res.sendFile(__dirname + '/views/beers.hbs')" before the API call and the render of the HBS view. That's the main problem.

PR #2344, opened 2022-07-06
Jack Le & Darren Iyoyo

jackletran:
done mandatory iterations with Darren

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,6 +1,7 @@
 const express = require('express');
 
 const hbs = require('hbs');
+const { get } = require('http');
 const path = require('path');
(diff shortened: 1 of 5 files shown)


app.js
```diff
@@ -1,6 +1,7 @@
 const express = require('express');
 
 const hbs = require('hbs');
+const { get } = require('http');
```
TA-Remote, line 4: Here you're importing the get from `http` but you are not using anywhere actually, if you take a look at your code it's not marked as been used and if you comment the line the pages keep working

TA-Remote (commented):
- Good work, but try to push ahead and go for the bonus as the partial concepts can be really useful to avoid repetition of your code

PR #2348, opened 2022-07-06
Lucas Borchardt & Paul Tiry [WDFT JUN2022]

ptiry02:
Finished without bonus. no time 😉


TA-Remote (commented):
- Good work! The styling is very good.

- Try work on the bonus when you have time to practice the partials concept too

PR #2349, opened 2022-07-06
Begum Eryildiz WDFTJune2022

begumeryildiz:
I solved the mandatory iterations together with Lara.
Then I also solved the bonus iterations.


TA-Remote:
Impressive Begum!

PR #2350, opened 2022-07-06
[WDFTJUN22]_lab_Ironbeers_Christoph_Bochniak_started_with_Nic_Solitom_no_Bonus

Christoph86:
all mandatory, no bonus, css,

mandatory iterations 1-4 together with Nic Solitom

### DIFF
--- a/views/beers.hbs
+++ b/views/beers.hbs
@@ -1,0 +1,10 @@
+<ul class="manyBeers">
+  {{#each beersFromApi}}
+    <li class="manyBeers">
+      <img src="{{this.image_url}}" alt="picture of {{this.name}}">
+      <h3>{{this.name}}</h3>
+      <h4 class="tagline">{{this.tagline}}</h4>   
+      {{this.description}} <br>
+    </li>
+  {{/each}}
+</ul>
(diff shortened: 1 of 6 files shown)


views/beers.hbs
```diff
@@ -0,0 +1,10 @@
+<ul class="manyBeers">
+  {{#each beersFromApi}}
+    <li class="manyBeers">
+      <img src="{{this.image_url}}" alt="picture of {{this.name}}">
+      <h3>{{this.name}}</h3>
+      <h4 class="tagline">{{this.tagline}}</h4>   
+      {{this.description}} <br>
+    </li>
```
TA-Remote, line 8: Keyword `this` is only necessary if your data  (object , array , whatever you are addressing) has no name to it.

In the case where you iterate over an array, and print info of every object,  in each iteration you are looking at one single object and can address it’s properties directly, so no need to use keyword `this`. But both syntaxes are valid

```
    <li class="manyBeers">
      <img src="{{image_url}}" alt="picture of {{name}}">
      <h3>{{name}}</h3>
      <h4 class="tagline">{{tagline}}</h4>   
      {{description}} <br>
    </li>
```

TA-Remote (commented):
Great work!

PR #2352, opened 2022-07-06
WD RT JUN 2022 | ruben Andino

rubenanlo:
done with bonuses except full format. I haven't added all the details for the single beers, but just because of the time.

Thanks in advance for your review!!


views/randomBeer.hbs
```diff
@@ -0,0 +1,13 @@
...
+        <ul>
+            {{#each this.food_pairing}}
+                <li>
+                    {{this}}
+                </li>
+            {{/each}}
+        </ul>
+    </p>
+    <p>Brewer's tips: {{this.brewers_tips}}</p>
+{{/each}}
```
TA-Remote, line 13: This is correct solution.
Another approach would be to skip keyword `this`, since you are already inside the beer object. So below syntax will give same result

```
{{#each responseFromApi}}
    {{>beerpartial}}
    <p>Food pairing:
        <ul>
            {{#each food_pairing}}
                <li>
                    {{this}}
                </li>
            {{/each}}
        </ul>
    </p>
    <p>Brewer's tips: {{brewers_tips}}</p>
{{/each}}
```

app.js
```diff
@@ -22,4 +23,29 @@ app.get('/', (req, res) => {
...
+    .catch(error => console.log(error))
+});
+
+app.get('/random-beer', (req, res) => {
+  punkAPI.getRandom()
+    .then(responseFromApi => {     
+      res.render('randomBeer',{responseFromApi});
+    })
+    .catch(error => console.log(error))
+});
```
TA-Remote, line 40: It is your preference, but my recommendation would be to format data in javascript, before you even send info to the view -> this way your view files will be short and easy to read.

So for example, on line 37, instead of sending an array with one element, you can send that single object directly

```
    .then(responseFromApi => {     
      res.render('randomBeer', responseFromApi[0]);
    })
```

TA-Remote (commented):
Ruben - awesome work 💪 

I had only suggestions, no corrections really 🙂

rubenanlo:
Thanks for the feedback!

PR #2354, opened 2022-07-07
Lara Rockenstein // WD FT June22 // mandatory iterations done together with Begüm

LMRBLN:


TA-Remote:
Great job Lara!

TA-Remote:
about iteration 6 : you need to write another route and receive id in that route as PARAMS. in this way you can show every beer detail. the route is like >>> 'beers/ :beerId' and you can have access to that inside request >>> request.params

app.get ('/beers/:beerId', (request, res) =>{
 const {beerId} = request.params
  punkAPI.getBeer(beerId)
    .then((beer) => {
      res.render('singleBeer',{beer})
    })

but dont worry :) you can see and learn this concept in the followinng days

PR #2512, opened 2022-11-09
[BER-FTWD-2022] Wei Geng

parrr-geng:


matigobbi (commented):
Good work Wei!
Nice uso of the .then, layout, and partials!
Keep going! ⭐

PR #2513, opened 2022-11-09
[BER - WDFT - NOV09] GaYeong Park

gayeongpark:


views/random-beer.hbs
```diff
@@ -0,0 +1,14 @@
+
+<div class = "random-beers">
+    <ul class="beer-list2">
+    {{#each responseFromAPI }}
+        <h2>{{this.name}}</h2>
+        <img src="{{this.image_url}}" alt="{{this.tagline}}">
+        <p>{{this.tagline}}</p>
+        {{#each this.food_pairing}}
+        <li>{{this}}</li>
```
matigobbi, line 9: be carefull with this line, you are printing the entire object, instead of refering any special Key-( like name, or image_url)

matigobbi (commented):
Good work GaYeong!!
Keep going!! :D :D

PR #2517, opened 2022-11-09
Berlin Oct2022 without css for bonus Ioannis Piastopoulos

ioannispiastopoulos:


matigobbi (commented):
Nice work Ioannis!
keep going!🚄

PR #2534, opened 2022-11-10
BER OCT2022 'done without styling' Anna Khablo

khablo1508:


thusspokedata (commented):
Nice home page. Everything is working perfectly! As an idea, you can use a horizontal card from bootstrap to do the beers page and random beer page.  Well Done!

PR #2563, opened 2023-01-12
1-5 including partials

JaenaJay:


JaenaJay:
I corrected my partials since it was having some issues but since I forgot to git add, and changed from punkAPI to Axios, I can't update my files any longer.

PR #2581, opened 2023-01-23
[WD PT NOV ES] - {Ana Gutiérrez Ruiz}

AnaGutierrezRuiz:


AnaGutierrezRuiz:
Lab not fully done, one bonus left;  still styles to implement and code to clean.

CrisArranz:
Nada de not fully done, esta de 10!

PR #2596, opened 2023-02-01
[FT-WD-Berlin-W4D3-lab] Indrajeet Patil

IndrajeetPatil:


views/random-beer.hbs
```diff
@@ -0,0 +1,9 @@
+{{#with beer}}
+
+  <div class='random-beer'>
+
+    {{> beerpartial this isRandom=true}}
+
+  </div>
+
+{{/with}}
```
mloelkes, line 9: Very good solution! 🙂

IndrajeetPatil: Thanks! :)

PR #2598, opened 2023-02-01
WDFT Paul Dombrowski - done without Bonus // Febr 2023

PaulDombrowski:


views/random-beers.hbs
```diff
@@ -0,0 +1,10 @@
+
+        <li>
+            {{#each beer}}
+            <img src={{image_url}}></br>
+            <h1>{{name}}</br></h1>
+            {{description}}</br>
+            <b>{{tagline}}</br></b>
+        </li>
+        {{/each}}
+
```
mloelkes, line 10: Good solutions so far! If you find the time, try to implement the partial for the beerCard. 🙂

app.js
```diff
@@ -19,7 +19,33 @@ app.use(express.static(path.join(__dirname, 'public')));
...
-  res.render('index');
+  res.render('index')
 });
 
+app.get('/beers', (req, res) => {
+  punkAPI
+    .getBeers()
+    .then(beersFromApi => {
+      let beers = [];
+      beersFromApi.forEach(beer => {
```
mloelkes, line 30: You don't need to push the beers to a new array, you can also just use the beersFromApi-array.

PR #2599, opened 2023-02-01
FTWD_Berlin_Januar_2023_IronBeers_OMA 

OfkirMohamed:


views/beers.hbs
```diff
@@ -0,0 +1,62 @@
+<!DOCTYPE html>
```
mloelkes, line 1: If you use a layout-file, you don't need to add the HTML-structure to the beers-view.

views/randombeers.hbs
```diff
@@ -0,0 +1,64 @@
...
+            <ul>
+                {{#each rbeer.food_pairing}}
+                    <li>{{this}}</li>
+                {{/each}}
+            </ul>
+            <div class="tips"> <strong>Brewers Tips:</strong> {{rbeer.brewers_tips}} </div>
+        </div>
+    </div>
+</body>
+</html>
```
mloelkes, line 64: Very good solutions! 👍

PR #2603, opened 2023-02-01
[WDFT Berlin] Alexis Aguilera

arekiu:


matigobbi (commented):
Good work Alex, is very clean in general, 
in lines 17, 25 and 32, you don't need to store the promise into a variable, mostly because you are not using in any another part then, but its not wrong. 
keep working ! 👯

PR #2642, opened 2023-02-08
[LISFT0123] João Carneiro

joao621119:


views/partials/beerpartial.hbs
```diff
@@ -0,0 +1,7 @@
+<div>
+
+     <img src="{{this.image_url}}" alt="beerimg">
+     <p><a href="/beers/beer-{{this.id}}">{{this.name}}</a></p>
```
webLisbonTA, line 4: This anchor tag with the name is displaying in white because you select all your `a` tags to be color: white in the CSS file.
Since the background of your website is white. It looks like there's no name of the beer, and nowhere to select the beer details page.

webLisbonTA (commented):
Good job on the lab João!

PR #2650, opened 2023-02-08
[LISFT0123] - Caroline Kuhn

Carolinekuuhn:


app.js
```diff
@@ -14,12 +14,38 @@ app.use(express.static(path.join(__dirname, 'public')));
...
+  punkAPI
+    .getBeers()
+    .then(beersFromApi => res.render('beers', { beersFromApi }))
+    .catch(error => console.log(error));
+});
+
+app.get('/random-beer', (req, res) => {
+  punkAPI
+    .getRandom()
+    .then(responseFromAPI => res.render('random-beer', { responseFromAPI }));
```
lzaquine, line 36: Don't forget to catch the error!

`.catch(error => console.log(error));`

app.js
```diff
@@ -14,12 +14,38 @@ app.use(express.static(path.join(__dirname, 'public')));
...
+});
+
+app.get('/beer-detailed', (req, res) => {
+  res.render('beer-detailed');
```
lzaquine, line 40: This one isn't working because if you go to your beer-detailed page, you only have a #each beerApi and the partial you created. If you want to display that, you would need to use the punkAPI in here.
So, for example:

```
punkAPI
  .getBeers()
  .then(beerApi => res.render('beer-detailed' , {beerApi}))
```

lzaquine (commented):
Great job on the lab, Caroline!

PR #2651, opened 2023-02-08
[LISBFT0123] Tomás Belmar 

TomBelmar:
sorry i couldn't finish.

### DIFF
--- a/app.js
+++ b/app.js
@@ -14,12 +14,44 @@ app.use(express.static(path.join(__dirname, 'public')));
 
 // Register the location for handlebars partials here:
 
-// ...
+hbs.registerPartials(path.join(__dirname, 'views/partials'));
 
-// Add the route handlers here:
+// Add the route handlers here: 
 
 app.get('/', (req, res) => {
   res.render('index');
 });
+
+app.get('/beers', (req, res) => {
+  punkAPI
+    .getBeers()
+    .then((beersFromApi) => {
+      console.log('Beers from the database: ', beersFromApi);
+      res.render('beers', {beersFromApi});
+    })
+    .catch(error => console.log(error));
+})
+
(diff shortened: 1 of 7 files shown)


views/layout.hbs
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="/public/stylesheets/styles.css">
```
joaopdg, line 7: In the `app.js` file you already specified that every static elements that you will be using in your website (ex: images, files) will be located inside the `public` folder:

```
app.use(express.static("public"));
```

So when you need to reference the `path` of something that is inside the `public` folder you don't need to specify public in the path you just need to right the path after the public folder:

```
<link rel="stylesheet" href="/stylesheets/styles.css">
```

app.js
```diff
@@ -14,12 +14,44 @@ app.use(express.static(path.join(__dirname, 'public')));
...
+    .getRandom()
+    .then((responseFromAPI) => {
+      console.log('Random beer from the database: ', responseFromAPI)
+      res.render('random-beer', {responseFromAPI})
+    })
+  .catch(error => console.log(error));
+});
+
+app.get('/beer-detailed', (req, res) => {
+  res.render('beer-detailed')
```
joaopdg, line 46: There is no need for this route `/beer-detailed` you already have the route `/beers/:id` that is doing the work of rendering the beerDetailed view

joaopdg (commented):
Great job on the lab Tomás!

PR #2652, opened 2023-02-08
[LISFT0123] Déborah

deborahdelimaa:
only missing css i guess

### DIFF
--- a/app.js
+++ b/app.js
@@ -3,6 +3,7 @@ const express = require('express');
 const hbs = require('hbs');
 const path = require('path');
 const PunkAPIWrapper = require('punkapi-javascript-wrapper');
+const { getEnabledCategories } = require('trace_events');
 
 const app = express();
 const punkAPI = new PunkAPIWrapper();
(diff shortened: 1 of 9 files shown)


app.js
```diff
@@ -23,3 +24,31 @@ app.get('/', (req, res) => {
 });
 
 app.listen(3000, () => console.log('🏃‍ on port 3000'));
+
+app.get("/", (req, res) => {
```
lzaquine, line 28: You didn't need this route in here! There's one right before it that is making it work

app.js
```diff
@@ -23,3 +24,31 @@ app.get('/', (req, res) => {
...
+
+app.get("/randomBeer", (req, res)=> {
+  punkAPI.getRandom()
+  .then(responseFromAPI => res.render("randomBeer", {responseFromAPI}))
+  .catch(error => console.log(error))
+})
+
+
+app.get('/beer-detailed', (req, res) => {
+  res.render('beer-detailed')
```
lzaquine, line 46: This one isn't working because if you go to your beer-detailed page, you only have a #each beerApi and the partial you created. If you want to display that, you would need to use the punkAPI in here.
So, for example:

```
punkAPI
  .getBeers()
  .then(beerApi => res.render('beer-detailed' , {beerApi}))
```

lzaquine (commented):
Great job on the lab & bonus, Deborah!

PR #2655, opened 2023-02-08
[LISFT0123] André Lisboa

andrelisboapt:
lab solved with bonus(css on layout and homepage)


lzaquine (commented):
Great job on the lab and doing the bonus, Andre!

PR #2656, opened 2023-02-08
[FTWD0123]Farid

Farid-97:


views/beers.hbs
```diff
@@ -0,0 +1,8 @@
+
+<div>
+  {{#each beersFromApi}}
+{{>beerpartial}}
+  <a href="h"></a>
```
webLisbonTA, line 5: This href is probably here by mistake. After the work is done, give a look in your code and check for this type of occurrences. It's not breaking anything in your code, but it's a bad pratice.

webLisbonTA (commented):
Good job on the lab Farid.

PR #2658, opened 2023-02-08
[LISFT0123] Pedro Leal

PeterLoyal:


app.js
```diff
@@ -21,5 +21,43 @@ app.use(express.static(path.join(__dirname, 'public')));
...
+
+});
+
+app.get('/beers/beer-:id', (req,res) => {
+
+  const id = req.params.id
+  punkAPI
+  .getBeer(id)
+  .then(each => {
+    res.render('random-beer', {each});
```
joaopdg, line 53: Since you're reusing the `random-beer` page to render the beer details (which is correct) you need to pass the response from the API with the same name that you're referencing inside the `random-beer` page (responseFromAPI):

```
app.get('/beers/beer-:id', (req, res) => {
  const id = req.params.id;
  punkAPI
    .getBeer(id)
    .then(responseFromAPI => {
    })
    .catch(err => console.log(err));
});
```

joaopdg (commented):
Great job on the lab Pedro!

PR #2661, opened 2023-02-08
Eveline Coenegrachts 

evelinecoen:


views/random-beers.hbs
```diff
@@ -0,0 +1,25 @@
+
+
+
+<section>
+<div>
+    {{> randombeerpartial }}
+   <p><h1>{{food_pairing}}</h1> </p>
```
lzaquine, line 7: You're not displaying the food_pairing and brewer_tips here.
For it to work, you should've used the each loop here.

```
<section>
    <div>
    {{> randombeerpartial }}
        {{#each randomBeers}}
    <p><h1>{{food_pairing}}</h1> </p>
    <p><a>{{brewers_tips}}</a></p>
        {{/each}}
    </div>
</section>
```

views/partials/beerpartial.hbs
```diff
@@ -0,0 +1,16 @@
+<section>
+ <!--{{#each beers}}-->
+ <div>
+    <img src="{{image_url}}" alt="beer name: {{name}}" class="beer-image">
+ </div>
+<div>
+   <a href="/beers/beer-i3f4d34s34b">{{name}}</a>
```
lzaquine, line 7: This id, `i3f4d34s34b `, was just an example. You should've created the route for a detailed beer, a view for it and used the other method the API gives you for it to work. And here, you would pass the params you passed on the route. We'll learn more about that this week, and if you want, try doing it again after it!

lzaquine (commented):
Good job on the lab, Eveline! Check the comments I left for a better understanding!

PR #2663, opened 2023-02-09
[LISFT0123]Marisha Deroubaix

Deroubaix:


views/partials/beersPartial.hbs
```diff
@@ -0,0 +1,7 @@
+<div class='element-image'>
+  <img src='{{image_url}}' alt='' />
+</div>
+
+<h3 class='name'><a href='/beers/{{this.id}}'>{{name}}</a></h3>
```
joaopdg, line 5: This is correct but you're missing the `route` to `/beers/:id` in the `app.js` file.

Hint: for this route you will need to extract `id` from the `req.params` and use it with a different method from the `punkAPI` so you can receive just one specific beer from the API.

joaopdg (commented):
Great job on the lab Marisha!

PR #2667, opened 2023-02-09
[LISFT0123] Gabriel

gbrlcsgn:


views/random-beer.hbs
```diff
@@ -0,0 +1,11 @@
+<div class='randomBeer'>
+  {{#each responseFromAPI}}
+    
+        {{>beerPartials this}}
```
webLisbonTA, line 4: Here, you forgot to initialize your paragraph with `<p>`.

webLisbonTA (commented):
Good job on the lab Gabriel. If you want, do the bonuses to practice.

PR #2669, opened 2023-02-09
[LISFT0123] Mariana Franco

marianamedeirosfranco:
Lab almost solved.

### DIFF
--- a/app.js
+++ b/app.js
@@ -14,12 +14,41 @@ app.use(express.static(path.join(__dirname, 'public')));
 
 // Register the location for handlebars partials here:
 
-// ...
+hbs.registerPartials(path.join(__dirname, 'views/partials'));
 
 // Add the route handlers here:
 
 app.get('/', (req, res) => {
   res.render('index');
+});
+
+app.get('/beers', (req, res) => {
+  punkAPI
+    .getBeers()
+    .then(beersFromApi => {
+      /*     console.log('Beers from the database: ', beersFromApi) */
+      res.render('beers', { beers:beersFromApi });
+    })
+    .catch(error => console.log(error));
+});
+
+//TEST to get ID
+app.get('/beer/:id', (req, res) => {
+  const beerId = req.params.id;
+  punkAPI.getBeers(beerId).then(responseFromAPI => {
+    res.render('beer/id', {random:responseFromAPI });
+  });
+});
+
+app.get('/random-beer', (req, res) => {
+  punkAPI
+    .getRandom()
+    .then(responseFromAPI => {
+      console.log('Random Beer from the database: ', responseFromAPI);
+
+      res.render('random-beer', { responseFromAPI });
+    })
+    .catch(error => console.log(error));
 });
 
 app.listen(3000, () => console.log('🏃‍ on port 3000'));
--- a/views/index.hbs
+++ b/views/index.hbs
@@ -1,3 +1,7 @@
 <div>
+<img src="/public/images/beer.png" alt="beer">
 
+<ul class = "beerButtons">
+            <li> <a class="" href="/views/beers.hbs"> Check the Beers! </a> </li>
+            <li> <a class="" href="/views/random-beer.hbs"> Check a Random Beer </a> </li>
 </div>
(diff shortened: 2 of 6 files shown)


views/layout.hbs
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="/public/stylesheets/styles.css">
```
joaopdg, line 7: In the `app.js` file you already specified that every static elements that you will be using in your website (ex: images, files) will be located inside the `public` folder:

```
app.use(express.static("public"));
```

So when you need to reference the `path` of something that is inside the `public` folder you don't need to specify public in the path you just need to right the path after the public folder:

```
<link rel="stylesheet" href="/stylesheets/styles.css">
```

views/index.hbs
```diff
@@ -1,3 +1,7 @@
 <div>
+<img src="/public/images/beer.png" alt="beer">
```
joaopdg, line 2: Same thing here, if you put `/public` the image will not be displayed:

```
<img src="/images/beer.png" alt="beer">
```

views/random-beer.hbs
```diff
@@ -0,0 +1,7 @@
+{{#each random}}
```
joaopdg, line 1: In the `app.js` file on the `/random-beer` route when you are rendering the `random-beer.hbs` you are also passing the response from the API as `responseFromAPI`:

```
```

So if you want to display the information that you get from the API you need to reference it with the same name as you're passing in the route:

```
{{#each responseFromAPI}}

  {{> beerpartial}}
  <h3>{{this.food_pairing}}</h3>
  <h3>{{this.brewer_tips}}</h3>

{{/each}}
```

views/index.hbs
```diff
@@ -1,3 +1,7 @@
...
 
+<ul class = "beerButtons">
+            <li> <a class="" href="/views/beers.hbs"> Check the Beers! </a> </li>
```
joaopdg, line 5: The `path` that you need to specify in the `href` is the `url` path that you specified in the `app.js` and not the path to the `html` file:

```
            <li> <a class="" href="/beers"> Check the Beers! </a> </li>
            <li> <a class="" href="/random-beer"> Check a Random Beer </a> </li>
```

app.js
```diff
@@ -14,12 +14,41 @@ app.use(express.static(path.join(__dirname, 'public')));
...
+      res.render('beers', { beers:beersFromApi });
+    })
+    .catch(error => console.log(error));
+});
+
+//TEST to get ID
+app.get('/beer/:id', (req, res) => {
+  const beerId = req.params.id;
+  punkAPI.getBeers(beerId).then(responseFromAPI => {
+    res.render('beer/id', {random:responseFromAPI });
```
joaopdg, line 39: You're using `punkAPI.getBeers` which is the same method that you used in the `/beers` route and this method returns you all the beers. But you just want one specific beer, so in the lab instruction they tell you that you have another method for this that is `punkAPI.getBeer(id)`:

```
  punkAPI.getBeer(beerId).then(responseFromAPI => {
```

Since here you want to render just one specific beer you can reuse the `random-beer.hbs` page and pass the response from the API with the same name:

```
```

views/partials/beerpartial.hbs
```diff
@@ -0,0 +1,4 @@
+<a href='/beers/{{id}}'></a><h1> {{name}}</h1>
+<img src='{{image_url}}' alt='beer' />
+<p> {{description}} </p>
+<p> {{tagline}} </p>
```
joaopdg, line 4: Here the link that you are giving to the `<a>` tag is `/beers/{{id}}` but the path that you specified in the routes id `/beer/{{id}}` also you're `<a>` is empty so the link will not work, to make it work you wrap it around all the tags so in this way no meter where you click you will go to the beer details page:

```
<a href='/beer/{{id}}'>

  <h1> {{name}}</h1>
  <img src='{{image_url}}' alt='beer' />
  <p> {{description}} </p>
  <p> {{tagline}} </p>

</a>
```

joaopdg (commented):
Hi Mariana, if you make the corrections that I pointed you'll be able to see your website fully working. Overall, good job so far!

PR #2679, opened 2023-03-15
[FTRMT022023] Marin - David

valyy151:
Iterations 1-5 complete

### DIFF
--- a/app.js
+++ b/app.js
@@ -12,36 +12,39 @@ app.set('views', path.join(__dirname, 'views'));
 
 app.use(express.static(path.join(__dirname, 'public')));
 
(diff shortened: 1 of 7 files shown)


TA-Remote:
Well done🙌!!!! 
As you may have noticed, not all beers have an image. It can happen often when working with api… Also in handlebars we can use if or if-else statements to display elements based on different data. 

TA-Remote:
Well done🙌!!!! 
As you may have noticed, not all beers have an image. It can happen often when working with api… Also in handlebars we can use if or if-else statements to display elements based on different data. 

PR #2680, opened 2023-03-15
update instructions to run with nodemon

luisjunco:


ross-u (changes requested):
@luisjunco Thanks for submitting the PR. 
Could we also add an additional note to the new FAQ right after the `npm install -g nodemon` code block, saying: "This command will install `nodemon` globally on your system, making it available to all of your projects."?

PR #2682, opened 2023-03-15
[wd.rmt.ft.feb-23] Peng Chen - Sotiris Ntokas

SotDok:
finished all iterations except half of bonus 6 :)

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,3 +1,4 @@
+const { response } = require('express');
 const express = require('express');
 
 const hbs = require('hbs');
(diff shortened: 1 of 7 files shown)


TA-Remote:
@SotDok @Mr-Chennn
well done guys!!! 
about the link: next step would be to build another route /beers/:id --> this way you can pass whatever is after the " : " dynamically. To get the value of the id you pass, you need to use req.param —> const myId = req.param so you can make a request to the punkApi using the getBeer(“passing the id in here —> myId”).  After, you can go in two ways: 
 1. Create another .hbs to render with the data of your single beer.
2. Go DRY and reuse the random-beer page that display the same data
Anyway, you'll see and use this pattern a lot soon😉

PR #2685, opened 2023-03-15
[RMT-WD-FEB-2023] Èric Vega

ericveganoguera:


TA-Remote:
Great job! Even with fever you're producing very good quality code🙌
As you may have noticed, not all beers have an image. So the the link is not always available.  Also in handlebars we can use if or if-else statements to display elements based on different data.  Just a small thing that can come very helpful 😉

PR #2686, opened 2023-03-15
[WD.FT.RMT.02.2023] Amina -Nihat

nihatalgan:
done without bonuses

### DIFF
--- a/app.js
+++ b/app.js
@@ -22,4 +22,44 @@ app.get('/', (req, res) => {
   res.render('index');
 });
 
+app.get('/beers', (request, response, next) => {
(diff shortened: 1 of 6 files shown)


ykztenoks:
Hi Amina && Nihat!
Congrats on your development! Great work on those iterations.
Small details that I could point out would be for when you render the beers, some beers don't have images, so instead of rendering the broken image with the alt property, you could add an if statement to check if the image exists, if not, you could render a <p> saying like "no image provided" or something like that.
Just a small detail, you nailed the iterations you did, keep it up!

PR #2687, opened 2023-03-15
[rmt-ft-wd-0223] Bartosz & Pim

8lom:
the random beer didn't show with the partial.

We were happy until Allesandra told us to do the bonus with the partial

### DIFF
--- a/app.js
+++ b/app.js
@@ -12,52 +12,35 @@ app.set('views', path.join(__dirname, 'views'));
 
 app.use(express.static(path.join(__dirname, 'public')));
 
-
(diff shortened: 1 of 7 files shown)


ykztenoks:
Hi Bartosz && Pim!
Great work on that lab! Nailed the iterations so far.
For the partial to work you'd need an extra page file for a single beer to then use the partials for the rendering. Your partial and randomBeer files looks totally fine you might just wanna check on the data you're receiveng to better handle it.
But you guys did a great job, keep it up!

PR #2689, opened 2023-03-15
all iterations and bonus done by both

andrehcarvalho6:


ykztenoks:
Hey Andre and Alex!
Awesome work, you guys nailed the iterations!
Clean, concise, great readability.
Nothing I could say but the usual, just clean up the code, delete dead code and commented lines, also, one thing would be to delete the `next` from the route arguments if you don't need it.
Keep it up!

PR #2690, opened 2023-03-15
[RMT-FT-WD FEB2023] Tuna & Chris 

TunaErkmen:


TA-Remote:
@TunaErkmen @Webster312
Well done!! I can see you tried to build different routes passing different properties of the .getBeers method result. Remember, one route can render one page (so one hbs file), Of course you can do multiple requests to retrieve different data and then pass them to your page…. For now let’s keep it simple😉

PR #2691, opened 2023-03-15
[RMT WD-FT FEB-2023] Cherry Leung & Ismael Freitas

IsmaelFreitas98:
We implemented all iteration, but cheated a bit in the partial part (Because of the different styling between the beer content in the beers page and the random beer page) .

### DIFF
--- a/app.js
+++ b/app.js
@@ -13,7 +13,7 @@ app.set('views', path.join(__dirname, 'views'));
 app.use(express.static(path.join(__dirname, 'public')));
 
 // Register the location for handlebars partials here:
-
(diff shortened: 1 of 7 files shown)


ykztenoks:
Hey Ismael && Cherry!
Awesome job on those iterations, you guys nailed it! 
Small detail that I could point out would be for when you render the beers, some beers don't have images, so instead of rendering the broken image with the alt property, you could add an if statement to check if the image exists, if not, you could render a p tag saying like "no image provided" or something like that.
Just a small detail, you rocked the iterations, keep it up!

PR #2692, opened 2023-03-15
[RMT FT WD FEB2023] Hugo & Mattie

maitlandbarton:
worked with Hugo for all the mandatory iterations, did styling & bonus 5 by myself

### DIFF
--- a/app.js
+++ b/app.js
@@ -14,7 +14,7 @@ app.use(express.static(path.join(__dirname, 'public')));
 
 // Register the location for handlebars partials here:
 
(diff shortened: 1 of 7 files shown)


TA-Remote:
Well done🙌!!!! 
As you may have noticed, not all beers have an image. It happens often also with other api.  Even in handlebars we can use if or if-else statements to display elements based on different data. 

`{{#if this.img_url}}
beer image
{{else}}
<p> sorry no image available</p>
{{/if}}
`

PR #2693, opened 2023-03-15
[RMT-FT-FEB-2023] Paulius Mockevicius - Casandra

pmockevicius:
We did iterations 1 - 5 together and then we started working on lab individually. Unfortunately had some family stuff going on and didn't have chance to work much on iteration 6 and only little time for iteration 7. Will be coming back to these in next few days

### DIFF
--- a/app.js
+++ b/app.js
@@ -14,7 +14,7 @@ app.use(express.static(path.join(__dirname, 'public')));
 
 // Register the location for handlebars partials here:
 
-// ...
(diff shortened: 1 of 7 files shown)


TA-Remote:
well done Paulius🙌!!
You'll have a lot of other opportunities to practice, especially the bonus you couldn’t complete😉 a small suggestion: when using img tags, get the habits to fill in the alt attribute (dynamically every time you can) with other properties related. It’s gonna turn very helpful later (when if you don’t have, it will complain🙄)

PR #2694, opened 2023-03-16
[RMTFTWD-FEB2023]Hugo and Mattie

hugowannabedev:
We completed the mandatory iterations

### DIFF
--- a/app.js
+++ b/app.js
@@ -22,4 +22,36 @@ app.get('/', (req, res) => {
   res.render('index');
 });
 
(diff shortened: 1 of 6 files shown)


TA-Remote:
Well done🙌!!!! 
As you may have noticed, not all beers have an image. It can happen often when working with api… Also in handlebars we can use if or if-else statements to display elements based on different data. 
`
{{#if this.image_url}} // check if we have data inside the property
 sorry, no image </p> // property empty 
{{/if}}
`

PR #2695, opened 2023-03-16
[RMT FT 20-FEB-2023]Hristos Pappas

hristosppp:


TA-Remote:
Well done🙌!!!! 
As you may have noticed, not all beers have an image. It can happen often when working with api…
 Also in handlebars we can use if or if-else statements to display elements based on different data.
`
{{#if this.image_url}} // check if we have data inside the property
 (or whatever you want) for a better user experience or avoid errors
{{/if}} ` 

PR #2696, opened 2023-03-16
[FT RMT 02202023] Cristian Palao

Dookietrukie:
No bonuses today because dentist took a while and got home late 😢 Will implement them over the weekend!

### DIFF
--- a/app.js
+++ b/app.js
@@ -33,14 +33,31 @@ app.get('/beers', (req, res, next) => {
         beers: beersRetrieved
       }
 
-      console.log(data.beers[1])
-
       res.render('beers', data);
 
     })
     .catch((error) => {
       console.log(error);
     });
+})
+
+// GET /random-beer
+app.get('/random-beer', (req, res, next) => {
+
+  punkAPI.getRandom()
+    .then((retrievedBeer) => {
+
+      const data = {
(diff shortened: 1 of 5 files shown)


TA-Remote:
Well done🙌!!!!
Annoying dentist😬… but you’ll have plenty of opportunity to practice! 
suggestion (that you might like):
object destructuring when passing data in your .hbs— > 

// GET /beers

// option 1 —> store it in a variable, helpful if you need to reuse that variable 
  punkAPI.getBeers()
    .then((beersRetrieved) => {
      const data = {
        beers: beersRetrieved
      }
      res.render('beers', data);
    })
    .catch((error) => {
      console.log(error);
    });
})

// option 2 —> directly when passing data
  punkAPI.getBeers()
    .then((beersRetrieved) => {  
      res.render('beers', {beers: beersRetrieved} ); // —> you’ll use beers property in your hbs

// option 3 —> directly when passing data calling the property as data you retrieve 
// it will assign the value beers(from DB) to an obj property with the same name 
// my fav so I don't have to rename things 😂 
  punkAPI.getBeers()
    .then((beers) => {  
      res.render('beers', {beers} ); // —> you’ll use beers property in your hbs

PR #2708, opened 2023-03-30
[RMT-PT-012023] Wiebke

fraulueneburg-ironhack:
**Some quick notes:**
- submitted two days later (as discussed with Shaun)
- Almost solved bonus 2 (iteration 6), but not quite


fraulueneburg-ironhack:
Almost solved bonus 2 (iteration 6), but not quite

PR #2767, opened 2023-05-03
[Lis-apr-23-wd-ft-Robson] lab-ironbeers

RobBatista:


views/beers.hbs
```diff
@@ -0,0 +1,8 @@
+<div>
+    {{#each beersFromApi}}
+        <img src="{{this.image_url}}" alt="beers">
```
brunorocha20, line 3: Since you are receiving from the API a string with the image URL, you shouldn't use double quotes to get the image.

views/randomBeer.hbs
```diff
@@ -0,0 +1,8 @@
+<div>
+    {{#each beersFromApi}}
+        <img src="{{this.image_url}}" alt="randomBeer">
```
brunorocha20, line 3: Like on the `beers` page, you are receiving from the API a string with the image URL, you shouldn't use double quotes to get the image.

views/beers.hbs
```diff
@@ -0,0 +1,8 @@
...
+        {{this.name}}
```
brunorocha20, line 4: And you need to pass the information that you are receiving, inside `<p>` or `<h1>` (paragraphs or heading tags).

brunorocha20 (commented):
Good job on the lab.

PR #2774, opened 2023-05-03
WD LIS APR 23 - Lucas Telles

lltelles:


views/beers.hbs
```diff
@@ -0,0 +1,8 @@
+<div>
+  {{#each beersFromApi}}
+    <h1>{{this.name}}</h1>
+    <p>{{this.tagline}}</p>
+    <p>{{this.description}}</p>
+    <img id="beers-images" src="{{this.image_url}}"
```
brunorocha20, line 6: Here, since the response from the API is retrieving a string with a link to the image, you should not pass the `this.image_url` inside double quotes.

brunorocha20 (commented):
Good job on the lab

PR #2775, opened 2023-05-03
[LIS-WDFT-APR23] João Carrilho

ghost:


views/beers.hbs
```diff
@@ -0,0 +1,21 @@
...
+    <link rel="stylesheet" href="/stylesheets/styles.css">
+    <title>Beers</title>
+</head>
+<body>
+<div>
+  {{#each beersFromApi}}
+    <h3>{{this.name}}</h3>
+    <p>{{this.tagline}}</p>
+    <p>{{this.description}}</p>
+    <img src='{{this.image_url}}' alt='' />
```
brunorocha20, line 16: Here, since the response from the API is retrieving a string with a link to the image, you should not pass the `this.image_url` inside double quotes.

brunorocha20 (commented):
Good job on the lab.

PR #2782, opened 2023-05-03
[WDFT LIS APR23] Victoria Avelar

avelarvicky:


views/beers.hbs
```diff
@@ -0,0 +1,19 @@
...
+    <link rel="stylesheet" href="/stylesheets/styles.css" />
+    <title>Beers</title>
+  </head>
+  <body>
+    {{#each beersFromApi}}
+        <img src="{{this.image_url}}" alt="">
+        <h1>{{this.name}}</h1>
+        <p>{{this.description}}</p>
+        <p>{{this.tagline}}</p>
+        {{> beerpartial this}}
```
brunorocha20, line 15: Here, you are duplicating your code. Since you are calling the partial and the information is coming from it, there's no need to leave the `image_url`, `name`, etc.

brunorocha20 (commented):
Good job on the lab.

PR #2787, opened 2023-05-04
[lis-wd-apr23]Niroj

Niroj167:


views/randomBeer.hbs
```diff
@@ -0,0 +1,8 @@
+<div>
+    {{#each beersFromApi}}
+        <img src="{{this.image_url}}" alt="beers">
+        {{this.name}}
```
brunorocha20, line 4: You can't forget to put the information that you are receiving from the API, inside it's designed tags. Since `this.name`, `this.description` and `this.tagline` are strings, you need to insert the response inside some `<p> or <h1>` (paragraphs or heading tags).

brunorocha20 (commented):
Good job on the lab.

PR #2790, opened 2023-05-04
[lis-apr23-wdft] cris

Cris-k1:


app.js
```diff
@@ -22,4 +22,25 @@ app.get('/', (req, res) => {
...
 });
 
+app.get('/beers', (req, res)=> {
+punkAPI
+  .getBeers()
+  .then(beersFromApi => res.render('beers', {beersFromApi}))
+  .catch(error => console.log(error));
+});
+
+app.get('/random-beer', (req, res)=> {
```
joaopdg, line 32: Here you're calling the .getRandom() method 2 times, you only need to call it once:
```
app.get('/random-beer', (req, res) => {
  punkAPI
    .getRandom()
    .then(responseFromAPI => {
    })
});
```

joaopdg (commented):
Hi Cris, take a look at the correction that I left you below, and check the lab's solution to see what you could have done differently.

PR #2798, opened 2023-05-31
[FTRMT052023] Marco - Fabio

mk3346:
finished with mandatory iterations, continuing individually on bonus iterations

### DIFF
--- a/app.js
+++ b/app.js
@@ -18,8 +18,49 @@ app.use(express.static(path.join(__dirname, 'public')));
 
 // Add the route handlers here:
 
-app.get('/', (req, res) => {
-  res.render('index');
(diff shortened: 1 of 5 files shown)


TA-Remote:
hey @mk3346  & @FabioSM46
Very good job completing the lab!🙌

- Cool that you wanted to try different way to manipulate the data from the API getting only the values requested before sending it to the views, not common

Small tips: 
- Often, working with API’s, happens that not all the fields are complete… As you have noticed some beers don’t have an image. For a better UX, and especially to avoid errors (in the future in other applications), use an if or if-else to check if the property exist or is not empty, can make the difference.

Alessandra

PR #2802, opened 2023-05-31
[FTRMT052023] cédricRocher - gregTildis

gtildis:
all iterations done except bonuses

### DIFF
--- a/views/beer.hbs
+++ b/views/beer.hbs
@@ -1,0 +1,5 @@
+<div>
+  <div class='beer-container'>
+   {{> beerpartial this}}
+  </div>
+</div>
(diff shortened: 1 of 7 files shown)


views/beer.hbs
```diff
@@ -0,0 +1,5 @@
+<div>
+  <div class='beer-container'>
+   {{> beerpartial this}}
```
TA-Remote, line 3: I would use camelCase for that one as it's a little bit more aligned with conventions than a 2 word file and all lower case

TA-Remote (commented):
- Amazing work, keep going
- Just remove the commented code before pushing to GitHub

PR #2803, opened 2023-05-31
RT-FT-WD05.23 Paula Mateo and Jeremy Couderc

JCouderc87:


app.js
```diff
@@ -22,4 +24,29 @@ app.get('/', (req, res) => {
...
+        res.render('beers', data);
+      })
+      .catch((err) => console.log("Error fetching one beer", err))
+    });
+
+    app.get('/random-beer', (req, res) => {
+      punkAPI.getRandom()
+        .then((responseFromAPI) => {
+          const data = {
+            randomBeer: responseFromAPI
```
TA-Remote, line 44: Here I would pass the element and not the whole array, that way you could skip the `#each` on the random-beer.hbs

```suggestion
            randomBeer: responseFromAPI[0]
```

views/random-beer.hbs
```diff
@@ -0,0 +1,8 @@
+{{#each randomBeer}}
+{{> beerpartial}}
```
TA-Remote, line 2: I would use camelCase here as we have a filename with 2 words

TA-Remote (commented):
- Good work, keep practicing
- Pay attention to the indentation of your code as it will make easier to follow it

PR #2804, opened 2023-05-31
[FTRMT052023] gokceBerndt / juanLozano

Gokceber:
It was fun. Thank you Ale and Daniel:)

### DIFF
--- a/app.js
+++ b/app.js
@@ -22,4 +22,32 @@ app.get('/', (req, res) => {
     res.render('index');
 });
 
-app.listen(3000, () => console.log('🏃‍ on port 3000'));
+app.get('/beers', (req, res) => {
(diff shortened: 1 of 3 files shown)


TA-Remote:
hey @Jualolo & @Gokceber 
Very good job with your iterations!🙌

Small tips: 
- Use console.log (they are free😜) to see the data structure of the data we are getting back from the Db or the API.  food_pairing is an array of strings, maybe you want to iterate with #each to display it as a list
- Often, working with API’s, happens that not all the fields are complete… As you have noticed some beers don’t have an image. For a better UX, and especially to avoid errors (in the future in other applications), use an if or if-else to check if the property exist or is not empty, can make the difference.

Alessandra

PR #2805, opened 2023-05-31
[FTRMT052023]Philippe Miossec - Linh Nguyen

linhvnde:
Iteration 1-7: done together


TA-Remote (commented):
- Amazing lab, keep practicing
- I would have used camelCase for the `beerPartial` (instruction should be updated)

PR #2807, opened 2023-05-31
[FTRMT052023] Carolin Goerner Lucas Gaviao

carolingoerner:


TA-Remote:
Hey @carolingoerner & @Lucas-Gaviao 
Very good job with your iterations and nice style🙌
Well done using obj destructuring passing the dat from the API in random-beer!

Small tips: 
- food_pairing is an array of strings, maybe you want to iterate with #each to display it as a list
- Often, working with API’s, happens that not all the fields are complete… As you have noticed some beers don’t have an image. For a better UX, and especially to avoid errors (in the future in other applications), use an if-else to check if the property exist or is not empty can make the difference.

Alessandra

PR #2808, opened 2023-05-31
FTRMT0523 - Eirik and I worked together for all major Iterations until Bonus 6/7 but pushing separately. 

jani500:


views/random-beer.hbs
```diff
@@ -0,0 +1,11 @@
+{{#each beer}}
+{{> beerpartial}}
```
TA-Remote, line 2: I would use camelCase here as it's more aligned with convention than having a 2 words file all lower case

app.js
```diff
@@ -22,4 +24,56 @@ app.get('/', (req, res) => {
...
+      console.log('error', error);
+    });
+});
+
+app.get('/random-beer', (req, res) => {
+  punkAPI
+    .getRandom()
+    .then(responseFromAPI => {
+      const data = {
+        beer: responseFromAPI
```
TA-Remote, line 47: I would pass just the element, so you could skip the `#each` in the render view

```suggestion
        beer: responseFromAPI[0]
```

TA-Remote (commented):
- Good work, keep practing :)

PR #2809, opened 2023-05-31
FTrmt052023 Mery and Nishtha have worked on code

nishag-26:


app.js
```diff
@@ -22,4 +22,20 @@ app.get('/', (req, res) => {
...
+    //const data = { beerAll: oneBeer[0]} 
+    console.log(beerAll)
+    res.render('beers',{beerAll: beerAll});
+  })).catch((error) => console.log(error))
+});
+app.get('/random-beer' ,(req, res) => {
+  punkAPI.getRandom()
+  .then((beer1) => {
+    console.log(beer1)
+    res.render('random-beer',{beer1: beer1});
```
TA-Remote, line 38: Here I would pass just the element, instead of the whole array, that way you could access the properties directly instead of using the `#each`.

I wouldn't call the property beer1, maybe oneBeer or randomBeer could be options

```suggestion
    res.render('random-beer',{ randomBeer: beer1[0] } );
```

TA-Remote (commented):
- Good work, keep pushing forward
- Pay attention to the indentation of your code as it makes easier for everyone to understand

PR #2810, opened 2023-05-31
[FTRMT052023] Djamel & Emily

estevenson6:


TA-Remote:
hey @estevenson6 & @ProgrammingDj
Very good job with your iterations🙌

Small tips: 
- food_pairing is an array of strings, maybe you want to iterate with #each to display it as a list
- For both, random beer and single beer detail (beer by id) we want to display the same data, we can also reuse the random-beer.hbs 
- Often, working with API’s, happens that not all the fields are complete… As you have noticed some beers don’t have an image. For a better UX, and especially to avoid errors (in the future in other applications), use an if-else to check if the property exist or is not empty can make the difference.

Alessandra

ProgrammingDj:
Hello Allessandra,

Ok yes makes sense. Thx for your feedback…
I had a wine 🍷 after the lab 😂.

Kind regards

Djamel

TA-Remote ***@***.***> schrieb am Do. 1. Juni 2023 um 11:16:

PR #2811, opened 2023-05-31
[FTRMT052023] Nuno-Lina

nunojsmonteiro:


views/partials/beerpartial.hbs
```diff
@@ -0,0 +1,9 @@
+{{#each beers}}
```
TA-Remote, line 1: I would remove the `#each` helper from here and use it in the other views

views/randomBeer.hbs
```diff
@@ -0,0 +1,3 @@
+
+
+{{> beerpartial}}
```
TA-Remote, line 3: I would use camelCase as you have a file with 2 words

TA-Remote (commented):
- Good work, keep practicing :)
- The layout page should be in singular, that's why your layout wasn't working

PR #2814, opened 2023-05-31
[FTRMT052023] Pedro Paulo + Fernando Ramos

pedroxpaulos:


TA-Remote:
hey @pedroxpaulos  & @fndoRamos
Very good job completing the lab!🙌

- For a step forward (and type even less 😜) you can use object destructuring to pass the data from the API to the view, it will create a property ourBeer containing the data of ourBeer from the API (as always you’ll use that property in the view)

```
app.get('/beers/:beer', (req, res) => {
  punkAPI
    .getBeer(req.params['beer'])
    .then(ourBeer => {
      console.log(ourBeer);
      res.render('beers', {ourBeer} );   // instead of { finalBeerList: ourBeer }
    })
    .catch(err => console.log('Sorry, we are dry.')); // GREAT ERROR MESSAGE!
});
```

Alessandra

PR #2891, opened 2023-08-02
[LIS-FT-JULY23] Bruno Cunha

brcunha3000:
Lab done minus Iteration 6 and i have to re-do the css because its all messed up (only the home page is properly done)


webLisbonTA (commented):
Good job on the lab Bruno.

PR #2894, opened 2023-08-03
lsbFT072023

Aamir269:


app.js
```diff
@@ -9,17 +8,40 @@ const punkAPI = new PunkAPIWrapper();
...
+      console.log(error);
+      res.status(500).send('An error occurred while fetching beers.');
+    });
+    
+});
+
+app.get('/random-beer', (req, res) => {
+  punkAPI
+    .getRandom()
+    .then(responseFromAPI => {
```
brunorocha20, line 35: Aamir, this method returns a single random beer. If you `console.log()` the response, you'd see that, so, there's no need to use that `randomValue` neither that `beer` variable to keep the first index of the `responseFromApi`.
You should've done it this way:
```
app.get('/random-beer', (req, res) => {
  punkAPI
    .getRandom() // .getRandom() is the method provided by punkAPI
    .then((responseFromApi) => {
      res.render('random-beer', { beer: responseFromApi });
    })
    .catch((error) => console.log(error));
});
```

brunorocha20 (commented):
Good job Aamir.

app.js
Aamir269: ok thank you bruno

PR #2896, opened 2023-08-03
 [LISFTJULY23] Mariana Marques

mfilipa97:


jacobviv (commented):
God job with the lab Mariana!

views/random-beer.hbs
```diff
@@ -0,0 +1,29 @@
+<!DOCTYPE html>
+<html lang="en">
```
jacobviv, line 2: You only need the `<html>`, ` <head>` and `<body>` in the layout.hbs
The rest of the hbs files will be just the divs you need.

PR #2903, opened 2023-08-23
[FTRM082023]RobertoKyuchukov-AlexisParsatLacoste

Leyberoh:
Lab done, but we used GPT for the bonus iteration 6.

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,4 +1,5 @@
 const express = require('express');
+//const hbs = require('hbs');
 
 const hbs = require('hbs');
 const path = require('path');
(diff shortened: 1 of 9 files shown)


TA-Remote:
Hi @Leyberoh & @RobertoZ182

to address the weird behaviour you mentioned this morning:

You wanna check for each beer obj if it has an image_url property, as you are displaying the img (as all the other data of a beer) in your partial, the if statement should be only there (you can remove it from the other files, as in there you are saing: if there is the image_url --> display the partial):

```
// beerpartial.hbs 

<p>{{name}}</p>
{{#if image_url}}
    <img src="{{image_url}}" width="100px" alt="Image of a beer">
{{else}} // —> you can also add an 'else' alternative
    <p>No image available for {{tagline}}</p>
{{/if}}
<p>{{tagline}}</p>
<p>{{description}}</p>
```
Hope that the brief explanation, will be clear… but please let me know if you have any questions🙏
Good job anyway🙌!!!!!
Ale

PR #2997, opened 2024-03-22
BER Webdev PT 012024 Nil Erden

nil-e:
When I tried doing the lab I found out that the PunkAPI is down and it will be decommissioned soon as well. So I used another api to get product data. So i am not really displaying beers but it should follow the same logic.

### DIFF
--- a/app.js
+++ b/app.js
@@ -13,11 +13,8 @@ app.set('view engine', 'hbs');
 app.set('views', path.join(__dirname, 'views'));
 
 app.use(express.static(path.join(__dirname, 'public')));
-
-// Register the location for handlebars partials here:
 hbs.registerPartials(path.join(__dirname, 'views/partials'));
 
-// Add the route handlers here:
 fetch('https://dummyjson.com/auth/login', {
   method: 'POST',
   headers: { 'Content-Type': 'application/json' },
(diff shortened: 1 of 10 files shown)


app.js
```diff
@@ -7,19 +7,87 @@ const PunkAPIWrapper = require('punkapi-javascript-wrapper');
...
+  let productData;
+  let itemNames = [];
+  itemList
+    .then(item => {
+      const { data, ...rest } = item;
+      productData = data.products;
+      return productData;
+    })
+    .then(() => {
+      for (let i = 0; i < 25; i++) {
```
jacked97, line 44: 1. avoid for loop with let i =0 sync, here you have define the length everytime. (use better available options .forEach, for(const each of array)

app.js
```diff
@@ -7,19 +7,87 @@ const PunkAPIWrapper = require('punkapi-javascript-wrapper');
...
+        itemNames.push(productData[i]);
+      }
+      return itemNames;
+    })
+    .then(() => res.render('beers', { itemNames }))
```
jacked97, line 49: when you returning in previous .then() you can just receive it as parameter, there is no need of using itemNames variable to store it (unless needed) [Promise Chaining]

jacked97 (changes requested):

